PR #1, opened 2019-07-16
Add `closest` and `common`.

sgibb:
I renamed `match.closest` to just `closest` which is the relaxed replacement for `match` and `common` replaces `%in%`.

Follow-up from:
https://github.com/rformassspectrometry/Spectra/pull/45#issuecomment-511680248

@jorainer:
>     1. I found it a little unintuitive that `tolerance` should be of length 1 or length `table` - I would have expected it to be of length `x` instead. I guess that's not something that could be changed?

As you assumed it isn't easy to change (especially without introducing any slow loops). I think length of 1 is mostly used. So I would like to keep it that way.

>     2. `match.closest` requires the input to be increasingly sorted while `matchApprox` works with any ordering. For us that's no problem as `mz` and `rt` are always expected to be sorted.

That's because `findInterval` just works with sorted values. No problem for us. So I would favour speed over flexibility here.

> 
>     4. I observed this issue: .... [lower and upper boundary are treated differently]:

I fixed it and used your code as unit test.


R/matching.R
```diff
@@ -0,0 +1,140 @@
...
+#' argument that allows relaxed matching.
+#'
+#' @param x `numeric`, the values to be matched.
+#' @param table `numeric`, the values to be matched against. In contrast to
+#' [`match`] `table` has to be sorted in increasing order.
+#' @param tolerance `numeric`, accepted tolerance. Use `0` to get the
+#' same behaviour as [`match`]. Use `Inf` (default) to match to the closest
+#' value without restrictions. Could be of length one or the same length as
+#' `table`.
+#' @param duplicates `character(1)`, it is possible that many `x` match one
```
jorainer: Just a minor thing: I find it easier for the user if parameter descriptions are short eventually referring to `@description` or `@note`. Suggestion: move this lengthy description to `@note` and shorten the parameter description.

R/matching.R
```diff
@@ -0,0 +1,140 @@
...
+#'
+#' ## Some new values
+#' x <- c(1.11, 45.02, 556.45)
+#' y <- c(3.01, 34.12, 45.021, 46.1, 556.449)
+#'
+#' ## Using a single tolerance value
+#' closest(x, y, tolerance = 0.01)
+#'
+#' ## Using a value-specific tolerance accepting differences of 20 ppm
+#' closest(x, y, tolerance = x * ppm(20))
```
jorainer: Shouldn't that be `tolerance = y * ppm(20)`? - length of tolerance should be either 1 or equal to `length(table)`

R/matching.R
```diff
@@ -0,0 +1,140 @@
...
+#'
+#' ## Same using 50 ppm
+#' closest(x, y, tolerance = x * ppm(50))
```
jorainer: Same here: `tolerance = y * ppm(50)`

R/matching.R
```diff
@@ -0,0 +1,140 @@
...
+
+    if (!is.numeric(tolerance) || tolerance < 0)
+        stop("'tolerance' has to be a 'numeric' larger or equal zero.")
+
+    if (!is.numeric(nomatch) || length(nomatch) != 1L)
+        stop("'nomatch' has to be a 'numeric' of length one.")
+
+    tolerance <- rep_len(tolerance, length(table)) + sqrt(.Machine$double.eps)
+    duplicates <- match.arg(duplicates)
+    ntable <- length(table)
```
jorainer: You could move this 2 lines up and then have 
`tolerance <- rep_len(tolerance, ntable) + sqrt(.Machine$double.eps)`, saves you one `length` call.

R/ppm.R
```diff
@@ -0,0 +1,13 @@
...
+#'
+#' @param x `numeric`, ppm values.
+#'
+#' @return `numeric`: `x * 1e6`
+#'
+#' @author Sebastian Gibb
+#' @export
+#' @examples
+#' ppm(5)
+ppm <- function(x)x * 1e6
```
jorainer: I like the idea of the `ppm` function - that's something I need a lot. But first you have a mistake here, it's either `x * 1e-6` or `x / 1e6` and secondly, I would define it differently:

```r
ppm <- function(x, ppm = 0) x * ppm / 1e6
```

So users could get the ppm for their values directly. In fact, if you ask for the *ppm* for the values `x` it will return them. For the `closest` call above it would then be: `closest(x, y, tolerance = ppm(y, 20))`.

R/matching.R
sgibb: You are right. That is left over from my initial attempt to change that according to your `matchApprox` implementation.

R/matching.R
sgibb: As above.

R/ppm.R
sgibb: Stupid me! Indeed its `x * 1e-6`. Fixed. BTW: that's exactly the reason why I need that function!

Regarding your second suggestion: I somehow like the `x * ppm(5)` idea but to be honest it is just useful for absolute errors (and your solution may avoid `x / ppm(5)` vs `x * ppm(5)` errors). How does this look like (I don't like the `ppm(1, 20)` here but is an *absolute* tolerance often used?)?:

| | absolute | relative |
|---|---|---|
|mine| `closest(x, y, tol = ppm(20))`|`closest(x, y, tol = y * ppm(20))`|
|yours|`closest(x, y, tol = ppm(1, 20))`|`closest(x, y, tol = ppm(y, 20))`|

R/matching.R
sgibb: Thanks! Changed it.

R/ppm.R
jorainer: Honestly, I've never used the ppm in terms of absolute error - I'm always considering it as relative to the provided m/z. In that sense the parts-per-million makes also more sense to me, it's somehow like a percentage of something, so I would never use it without the *something* (which in our case is the m/z). AFAIK the ppm for MS instruments are also relative to the measured m/z - or am I wrong here?

I would like it to be changed.

sgibb: ok, done!

R/ppm.R
```diff
@@ -1,13 +1,14 @@
 #' PPM - Parts per Million
 #'
-#' `ppm` is a small helper function to mulitply *ppm* values with `1e-6`.
+#' `ppm` is a small helper function to mulitply *mz* values with a given *ppm*.
```
jorainer: Alternative formulation?

`ppm` is a small helper function to determine the parts-per-million for a user-provided value and *ppm*.

sgibb: Yes.

jorainer:
I'd say you can merge these changes (2 out of 3 core developers agree).

PR #5, opened 2019-07-19
Add logMessage

sgibb:
This PR adds the function `logMessage` that should be used in our other packages to allow a standardised log format.

Questions:

1. Different name? Before the function was named `logging` in `MSnbase`/`Spectra`, e.g. `logging`, `logmsg`, `logMsg`, `appendLogMessage`.
2. Different date format? Currently we use `date()` which yield `"Fri Jul 19 09:59:50 2019"`, what about `"2019-07-19 09:59:50"` instead? (24 vs 19 characters) Or even shorter versions, like `"20190719 095950"` or just a timestampe (isn't really readable by the user)?


lgatto:
Ok, so it is mostly a misunderstanding of the meaning of `...` - I interpreted it as several possible messages, you as a composition of one message. The reason I prefer to compose my message myself is that I can choose `paste` or `paste0`. With the current `...`, I'm sure I'll have to check whether `paste` or `paste0` is called inside `logMessage` - in other words, I prefer a little more typing as opposed to uncertainty. May be @jorainer has an opinion. 

As for `datestr`, I think the default should be to add a date. Hence have a `date = TRUE` argument will be the easiest in 99% or situations. But we would have a `datestr` function (callewd inside `logMessage` that we could reuse whenever there's a date to be composed).

sgibb:
@lgatto it may be a stupid suggestion but would renaming the function into `logMessage0` actually help? (similar to `paste0`, `cat0`, ...)

lgatto:
Not sure - my confusion stems rather from the meaning of `...`: one vs multiple message. May be the easiest would be that I provide some use cases and suggest an alternative implementation, and then we vote.

jorainer:
I don't have any strong opinion on the use or misuse of the `...` - actually, keeping the implementation as @sgibb suggested would satisfy both scenarios with and without using `paste0`.

lgatto:
I prefer option 2 (only) as it is explicit. I think having multiple ways of doing the same thing can lead to unnecessary confusion.

sgibb:
If we prefer
```r
object@processing <- logMessage(object@processing, 
                                paste0("Removed ", nRemove, " of ", nTotal " peaks"))
```

the whole function becomes useless because the following is nearly the same amount of typing (and you are doing all what the function is doing manually while defining its argument):
```r
object@processing <- c(object@processing, 
    paste0("Removed ", nRemove, " of ", nTotal " peaks [", date(), "]")
```
(as discussed above we could provide a `datestr <- function()paste0(" [", date(), "]")` to ensure the same date format in every log message)

The current interface is very similar to the one of `cat` (`cat <- function(..., file = "", sep = " ", fill = FALSE)`).

I don't think this function is critical in any way. That's why I close this PR in favour of a to be written `datestr` function and a `c(x, paste0)` statement.

PR #6, opened 2019-07-19
localMaxima

sgibb:
This PR adds `noise` and `localMaxima`. Both functions are useful for peak detection/centroiding. I am not sure that we really need them outside of `Spectra`. So feel free to ask me to move them to `Spectra` :wink:
BTW: both are taken more or less identically from `MALDIquant`.


jorainer:
👍  I do think they will be useful also for other packages - I'm using them already in `xcms`

R/localMaxima.R
```diff
@@ -0,0 +1,23 @@
+#' Local Maxima
```
jorainer: Some very minor thing that you can also ignore: I prefer to always use roxygen keywords, i.e. this would become `@title ...` and the description below `@description`. But as said, you can also simply ignore that suggestion.

sgibb: Ok, that's more descriptive. I will do that (but without opening a PR to review it).

PR #10, opened 2019-07-26
Add groupRun and joinNumeric

jorainer:
- `groupRun` represents the core of the `MSnbase:::.group_mz_values` function: this function groups ordered numerics if their difference is smaller than `tolerance + ppm`.
- `joinNumeric` uses `groupRun` to define which values from two numerics are close enough to be joined (accepting a `tolerance` and `ppm`). This function will be used in a function in `Spectra` to map/join peaks between two spectra if their m/z are close allowing different types of *joins* (outer, inner, left, right) - discussed in https://github.com/rformassspectrometry/Spectra/issues/46

Note that `joinNumeric` will also be helpful for a simple mapping/aligning/correlating chromatographic data.


sgibb:
Thanks for this PR. I really like the way `joinNumeric` is working. It is
very elegant. But IMHO there are a few major problems with the `groupRun`
function (that affect `joinNumeric`).

1. For multiple matches the **last** is reported. IMHO it should be the
**closest** one.
2. The `tolerance` has to be smaller than the gap between the groups
otherwise `groupRun` won't work and `joinNumeric` creates wrong results.
I guess this wouldn't be a real problem for centroided spectra:


```r
library("MsCoreUtils")
x <- seq(from = 0, to = 5, by = 0.1)
y <- 3:4
joinNumeric(x, y, tolerance = 0.2, join = "inner")
```

```
## Warning in joinNumeric(x, y, tolerance = 0.2, join = "inner"): Multiple
## matches between elements in 'x' and 'y'
```

```
## $x
## [1] 51
##
## $y
## [1] 2
#
# I would expect the following output
# x: 31, 41
# y: 1, 2
```

I proposed another PR #11 that IMHO solves both problems. What do you think
about it?

jorainer:
I agree on both your concerns:

> 1) For multiple matches the last is reported. IMHO it should be the closest one.

agree - only, for centroided spectra it should be a 1:1 mapping only and for that it works - if it stumbles across an 1:n mapping a warning is thrown. My thinking was that if users get a warning they should choose another mapping function (that can handle multi-mappings) - this function was supposed to be as fast as possible (which I think it is).

> The tolerance has to be smaller than the gap between the groups
otherwise groupRun won't work and joinNumeric creates wrong results.
I guess this wouldn't be a real problem for centroided spectra:

Agree. But that will be a problem for any mapping function. If the tolerance is larger than the distance of peaks within the same spectrum mapping will always fail.

I'll have a look at your PR.

PR #14, opened 2019-08-20
Add smooth

sgibb:
This PR adds the `smooth` function to `MsCoreUtils`.

I am not sure that my current approach is the best interface. For
`smooth(mz, hws = 3, method = "SavitzkyGolay")` half of the run time is needed
to calculate the filter coefficients for the *Savitzky-Golay-Filter*. These
coefficients don't depend on the input vector `x`. So for `Spectra` it would be
useful to calculate the coefficients once using `.coef*` and use `.filter` to
smooth all spectra.

Should we export `.coef*` and `.filter`? Or change the interface to `smooth`?


sgibb:
Now I removed the `method` argument. `smooth` takes two arguments: `x` and `cf` (coefficient matrix). While this may a bit awkward it greatly improves the runtime for more than one call of `smooth` because the coefficients are always identical (and calculation of it took from 1/10 (`coefMA`) to 1/2 (`coefSG`) of the whole runtime of `smooth`).

PR #19, opened 2019-09-05
Add `between` and `%between%`

sgibb:
This PR adds the `between` function and `%between%` operator to simplify tests
for values in ranges.

This e.g. improves the run time from `i2index`:

```r
library("MsCoreUtils")

f1 <- function(i, length) {
    if (is.integer(i) && !all(abs(i) %in% seq_len(length)))
        stop("index out of bounds: index has to be between 1 and ", length)
    i
}

f2 <- function(i, length) {
    if (is.integer(i) && !all(range(abs(i)) %between% c(1L, length)))
        stop("index out of bounds: index has to be between 1 and ", length)
    i
}

x <- seq(1e5)
n <- length(x)

all.equal(f1(x, n), f2(x, n))
# [1] TRUE

library(microbenchmark)
microbenchmark(f1(x, n), f2(x, n))
# Unit: microseconds
#      expr       min        lq       mean     median         uq       max neval
#  f1(x, n) 41756.140 43795.075 51091.9338 46760.9040 52877.5360 82769.828   100
#  f2(x, n)   585.753   616.648   757.4091   649.6625   806.0765  3192.329   100
```

Do think this is useful? Good name? (there is a similar function in `dplyr` of course)


lgatto:
Nice!

PR #20, opened 2019-10-03
graphPeaks for matching two spectra and returning a matched spectra

tnaake:
As discussed [here](https://github.com/rformassspectrometry/Spectra/issues/49), I have written a function `graphPeaks` that should be integrated into `MsCoreUtils`. Please note, I changed the input of dotproduct to be two matrices, actually this was discussed also in the conversation.


sgibb:
Wow, thanks for this PR. I will wait for #17 to be merged before starting to review this huge PR.

sgibb:
I am going to apply the following commands to undo all this accidentally merged  commits:
```
git checkout master
git reset --hard 4aabfe2
git push -f
```
It could be that some (e.g. @tnaake ) have to `git pull` the master again. But anyway we want to work on `devel` so it should not hurt to many people.

Another possibility would be to `git revert` every single commit. But I am afraid what would happen if we merge `devel` into `master` (both share some of the commits now, one would be reverted the other not). On the other hand the history would be very cluttered.

@lgatto and @jorainer  Are you satisfied with `git reset --hard 4aabfe2`?

BTW: I can't reopen a merged PR. @tnaake could please recreate a new PR against our `devel` branch?

sgibb:
I did a `git revert`. So `master` is on the same state as before the merging. We could just keep it as it is or to a potential dangerous `git reset --hard + git push -f` to keep our history clean.

What do you mean?

PR #21, opened 2019-10-12
Check and validation function for peaks matrices

sgibb:
This PR introduces the discussed `isPeaksMatrix` function. 

@jorainer: I know that this is handled by `Spectra` in general but if somebody uses `dotproduct` or similar functions outside of `Spectra` we have to test this, or?

I am not sure that we really need this function. @lgatto , @jorainer What do you think?


R/checks.R
```diff
@@ -0,0 +1,24 @@
...
+#' isPeaksMatrix(1:2)
+#' isPeaksMatrix(cbind(mz = 1:2, intensity = 1:2))
+isPeaksMatrix <- function(x) {
+    isTRUE(
+        is.matrix(x) &&
+        mode(x) == "numeric" &&
+        ncol(x) == 2L &&
+        !is.null(colnames(x)) &&
+        all(colnames(x) == c("mz", "intensity"))
+    )
```
jorainer: You might also want to check that m/z are ordered increasingly.

sgibb: Indeed. I will add that. In general I am a little bit afraid that all these checks increase our run time dramatically (especially in often called functions like `dotproduct`).

jorainer: Agree. That's why I would **not** call this function within `dotproduct`. But with this function we provide the user a tool to check input *before* calling `dotproduct`.

PR #24, opened 2019-10-20
Add a function `group`

jorainer:
This PR adds the function `group` that *groups* numeric values if their pairwise difference is smaller than `tolerance` and/or `ppm`. We had the same/similar functionality in `MSnbase`, this one works however also with unsorted `numeric` values and fixes floating point comparisons.

Ultimately I will reuse this function in the `combineSpectra` method in `Spectra`. This function is then used to group peaks from the set of input spectra (that's why it's different from the match/join functions we have already, that allow only pairwise matching).


DESCRIPTION
```diff
@@ -16,7 +16,7 @@ Authors@R: c(person(given = "Laurent", family = "Gatto",
                     role = c("aut", "cre"),
 		            comment = c(ORCID = "0000-0001-7406-4443")),
              person(given = "Sigurdur", family = "Smarason", role = "ctb"),
-	           person(given = "Thomas", family = "Naake", role = "ctb")
+	         person(given = "Thomas", family = "Naake", role = "ctb")
```
sgibb: Could you please fix this indentation (I thought I already did that, maybe just (re)merge with master).
```suggestion
	           person(given = "Thomas", family = "Naake", role = "ctb")
```

R/group.R
```diff
@@ -0,0 +1,71 @@
...
+#' ## Values 65, 65.1 and 65.2 have been grouped into the same group.
+group <- function(x, tolerance = 0, ppm = 0) {
+    if (is.unsorted(x)) {
+        idx <- order(x)
+        x <- x[idx]
+    } else idx <- integer()
+    xdiff <- diff(x)
+    tolerance <- tolerance + sqrt(.Machine$double.eps)
+    if (ppm > 0) {
+        res <- cumsum(c(0L, xdiff >= (tolerance +
```
sgibb: `cumsum(c(1L, xdiff ...)` is faster than `cumsum(c(0L, xdiff ...) + 1L`

R/group.R
```diff
@@ -0,0 +1,71 @@
...
+                                      ppm(x[-length(x)], ppm)))) + 1L
+    } else
+        res <- cumsum(c(0L, xdiff >= tolerance)) + 1L
+    if (length(idx))
+        res[order(idx)]
```
sgibb: The additional `order` call slows things down, why not use the index directly? I would rewrite the function as follows:
```r
group1 <- function(x, tolerance = 0, ppm = 0) {
    if (is.unsorted(x)) {
        idx <- order(x)
        x <- x[idx]
    } else
        idx <- integer()

    tolerance <- tolerance + sqrt(.Machine$double.eps)

    if (ppm > 0)
        tolerance <- tolerance + ppm(x[-length(x)], ppm)

    x <- cumsum(c(1L, diff(x) >= tolerance))
    x[idx] <- x
    x
}

set.seed(2019)
n <- 1e5
mz <- sample(n, n, replace = TRUE) + rnorm(n)
smz <- sort(mz)

microbenchmark(group(mz, ppm = 20), group1(mz, ppm = 20), check = "equal")
# Unit: milliseconds
#                  expr       min        lq     mean    median       uq      max
#   group(mz, ppm = 20) 10.436601 10.807413 11.59656 11.210812 12.14606 15.35732
#  group1(mz, ppm = 20)  8.423567  8.750131 10.16782  9.096824 10.10651 73.36924
#  neval
#    100
#    100
microbenchmark(group(smz, ppm = 20), group1(smz, ppm = 20), check = "equal")
# Unit: milliseconds
#                   expr      min       lq     mean   median       uq      max
#   group(smz, ppm = 20) 3.267696 3.421522 3.852874 3.536960 4.240001 7.588491
#  group1(smz, ppm = 20) 3.094932 3.245265 3.685689 3.321149 3.517340 7.728484
#  neval
#    100
#    100
microbenchmark(group(smz), group1(smz), check = "equal")
# Unit: milliseconds
#         expr      min       lq     mean   median       uq      max neval
#   group(smz) 2.398114 2.482195 2.897100 2.554832 2.785286 5.241478   100
#  group1(smz) 2.256011 2.304851 2.621905 2.385354 2.616277 5.361702   100
```

R/group.R
```diff
@@ -0,0 +1,67 @@
...
+#' ## Values 65, 65.1 and 65.2 have been grouped into the same group.
+group <- function(x, tolerance = 0, ppm = 0) {
+    if (is.unsorted(x)) {
+        idx <- order(x)
+        x <- x[idx]
+    } else idx <- integer()
+    tolerance <- tolerance + sqrt(.Machine$double.eps)
+    if (ppm > 0)
+        tolerance <- tolerance + ppm(x[-length(x)], ppm)
+    res <- cumsum(c(0L, diff(x) >= tolerance)) + 1L
```
sgibb:
```suggestion
    res <- cumsum(c(1L, diff(x) >= tolerance))
```

jorainer: I've seen that in your implementation - but is it still working? isn't adding `+1` at the end different than making the `cumsum` starting from `1`?

sgibb: >  isn't adding +1 at the end different than making the cumsum starting from 1?

Mathematically there is no difference but the later is slightly faster:

```r
library("microbenchmark")
cs1 <- function(x)cumsum(c(0L, x)) + 1L
cs2 <- function(x)cumsum(c(1L, x))

set.seed(2019)
x <- sample(c(TRUE, FALSE), 100, replace = TRUE)
microbenchmark(cs1(x), cs2(x), check = "identical")
# Unit: microseconds
#    expr   min     lq    mean median    uq     max neval
#  cs1(x) 1.442 1.6105 2.04819 1.7500 1.970  24.821   100
#  cs2(x) 1.092 1.2940 3.96220 1.4355 1.604 246.849   100

set.seed(42)
x <- sample(c(TRUE, FALSE), 100, replace = TRUE)
microbenchmark(cs1(x), cs2(x), check = "identical")
# Unit: microseconds
#    expr   min    lq    mean median     uq    max neval
#  cs1(x) 1.439 1.594 1.92930 1.6790 1.8645 21.955   100
#  cs2(x) 1.073 1.273 1.38073 1.3315 1.4575  2.695   100
```

jorainer: Did that now in the last commit.

PR #32, opened 2019-12-11
Adds the aggregation functions

lgatto:


DESCRIPTION
```diff
@@ -23,7 +23,9 @@ Depends:
     R (>= 3.6.0)
 Imports:
     methods,
-    S4Vectors
+    S4Vectors,
+    stats,
+    MASS
```
sgibb: I know `MASS` is a recommended package but would you mind to put it under `Suggests:` instead? (with `requireNamespace` in `robustSummary`)

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
+##' This function calculates the robust summarisation for each feature
+##' (protein). Note that the function assumes that the intensities in
+##' input `e` are already log-transformed.
+##'
+##' @title Return the Robust Expression Summary of a matrix
```
sgibb: I would prefer to have the title in the first line.

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##' @param x A feature by sample `matrix` containing quantitative
+##'     data with mandatory `colnames` and `rownames`.
+##' @param ... Additional arguments passed to [MASS::rlm()].
+##' @return `numeric()` vector of length `ncol(x)` with robust
+##'     summarised values.
+##' 
+##' @author Adriaan Sticker, Sebastian Gibb and Laurent Gatto
```
sgibb: Shouldn't Adriaan Sticker also be added as contributor to the `Authors@R` or at least to the `README.md`?

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##' 
+##' @importFrom MASS rlm
+##' 
+##' @examples
+##' x <- matrix(rnorm(30), nrow = 3)
+##' colnames(x) <- letters[1:10]
+##' rownames(x) <- LETTERS[1:3]
+##' robustSummary(x)
+robustSummary <- function(x, ...) {
+    stopifnot(!is.null(colnames(x)))
```
sgibb: Actually `stopifnot` is a wrapper around a lot of code and will throw a more or less useless error message to the user. Would it be better to add `if (!is.null(colnames(x)) stop("colnames must not be empty")` here and the row below?

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+
+    ## Sum contrast on peptide level so sample effect will be mean
+    ## over all peptides instead of reference level.
+    X <- stats::model.matrix(~ -1 + sample + feature,
+                             contrasts.arg = list(feature = 'contr.sum'))
+    ## MASS::rlm breaks on singulare values.
+    ## - Check with base lm if singular values are present.
+    ## - If so, these coefficients will be zero, remove this collumn
+    ##   from model matrix
+    ## - Rinse and repeat on reduced modelmatrx till no singular
```
sgibb:
```suggestion
    ## - Rinse and repeat on reduced modelmatrix till no singular
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+    ##   values are present
+    repeat {
+        fit <- stats::.lm.fit(X, expression)
+        id <- fit$coefficients != 0
+        X <- X[ , id, drop = FALSE]
+        if (!any(!id)) break
+    }
+    ## Last step is always rlm: calculate estimated effects effects as
```
sgibb:
```suggestion
    ## Last step is always rlm: calculate estimated effects as
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+    ## summarised values
+    fit <- MASS::rlm(X, expression, ...)
+
+    sampleid <- seq_along(unique(sample))
+    ## This will be needed for NUSE-type of quality control, but will
```
sgibb: Couldn't this comment be removed?

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+    res <- coef[paste0('sample', colnames(x))]
+    names(res) <- colnames(x)
+    res
+}
+
+##' Fits an additive model (two way decomposition) using Tukey's median
+##' polish procedure using [stats::medpolish()].
+##'
+##' @title Return the Median Polish (Robust Twoway Decomposition) of a matrix
+##' @param x A numeric matrix.
```
sgibb:
```suggestion
##' @param x A `matrix` of mode `numeric`.
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##' @param verbose Default is `FALSE`.
+##' @param ... Additional arguments passed to [stats::medpolish()].
+##' 
+##' @return A vector of numerics of length identical to `ncol(x)`.
```
sgibb:
```suggestion
##' @return A `numeric` vector of length identical to `ncol(x)`.
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##' @export
+##'
+##' @importFrom stats medpolish
+##' 
+##' @examples
+##' x <- matrix(rnorm(30), nrow = 3)
+##' medianPolish(x)
+medianPolish <- function(x, verbose = FALSE, ...) {
+    medpol <- stats::medpolish(x, trace.iter = verbose, ...)
+    return(medpol$overall + medpol$col)
```
sgibb:
```suggestion
    medpol$overall + medpol$col
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##'   [MASS::rlm()];
+##'
+##' - [base::colMeans()] to use the mean of each column;
+##'
+##' - [base::colSums()] to use the sum of each column;
+##'
+##' - [matrixStats::colMedians()] to use the median of each column.
+##'
+##' @title Aggreagate quantitative features.
+##' @param x A numeric matrix.
```
sgibb:
```suggestion
##' @param x A `matrix` of mode `numeric`.
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##' @param INDEX A factor of length `nrow(x)`.
```
sgibb:
```suggestion
##' @param INDEX A `factor` of length `nrow(x)`.
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##' @param FUN A function to be applied to the subsets of `x`.
```
sgibb:
```suggestion
##' @param FUN A `function` to be applied to the subsets of `x`.
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##' @param ... Additiona arguments passed to `FUN`.
```
sgibb:
```suggestion
##' @param ... Additional arguments passed to `FUN`.
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##'                                  "iTRAQ4.116", "iTRAQ4.117")))
+##' x
+##' 
+##' k <- factor(c("B", "E", "X", "E", "B", "B", "E"))
+##'
+##' aggregate_by_vector(x, k, colMeans)
+##' aggregate_by_vector(x, k, robustSummary)
+##' aggregate_by_vector(x, k, medianPolish)
+aggregate_by_vector <- function(x, INDEX, FUN, ...) {
+    stopifnot(inherits(x, "matrix"))
```
sgibb: As above in `robustSummary`:
```suggestion
    if (!inherits(x, "matrix"))
        stop("'x' has to be a 'matrix'.)
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+    stopifnot(identical(length(INDEX), nrow(x)))
```
sgibb:
```suggestion
    if (!identical(length(INDEX), nrow(x)))
        stop("The length of 'INDEX' has to be identical to 'nrow(x)'.)
```

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+##'   [stats::medpolish()];
+##'
+##' - [robustSummary()] to calculate a robust aggregation using
+##'   [MASS::rlm()];
+##'
+##' - [base::colMeans()] to use the mean of each column;
+##'
+##' - [base::colSums()] to use the sum of each column;
+##'
+##' - [matrixStats::colMedians()] to use the median of each column.
```
sgibb: Is this working in `R CMD check` without a warning? Even if `matrixStats` is not in `Suggests`?

tests/testthat/test_aggregate.R
```diff
@@ -0,0 +1,101 @@
...
+    expect_identical(length(x2), ncol(x))
+    mp <- stats::medpolish(x, trace.iter = FALSE)
+    ## Check decomposition
+    expect_identical(x, mp$overall + outer(mp$row, mp$col, "+") + mp$residuals)
+    expect_identical(names(x2), colnames(x))
+})
+
+
+test_that("aggregation: robustSummary", {
+    ## numeric example taken from `MSnbase::combineFeatures` on
```
sgibb: Do you want to check for the errors here as well (`is.null(colnames(x))`)?

R/aggregate.R
```diff
@@ -0,0 +1,166 @@
...
+    ## - Check with base lm if singular values are present.
+    ## - If so, these coefficients will be zero, remove this collumn
+    ##   from model matrix
+    ## - Rinse and repeat on reduced modelmatrx till no singular
+    ##   values are present
+    repeat {
+        fit <- stats::.lm.fit(X, expression)
+        id <- fit$coefficients != 0
+        X <- X[ , id, drop = FALSE]
+        if (!any(!id)) break
```
sgibb:
```suggestion
        if (all(id)) break
```

sgibb:
BTW: you not followed our commit convention 😜

lgatto:
Damn, forgot! :-(

R/aggregate.R
lgatto: Yes, at least on my computer. Will double check.

DESCRIPTION
lgatto: Do we need `requireNamespace` at all, given that it is always available and I call the function with the `MASS::` prefix?

R/aggregate.R
lgatto: Done

R/aggregate.R
lgatto: Done

R/aggregate.R
lgatto: Done

R/aggregate.R
lgatto: Done.

tests/testthat/test_aggregate.R
lgatto: Done

R/aggregate.R
lgatto: Done.

lgatto:
Updated the PR (fix properly formatted commit messages this time) to address all your points.

R/aggregate.R
sgibb: Ok, until now I also added the contributors to the `README.md` where I briefly described what they did.

R/aggregate.R
```diff
@@ -0,0 +1,175 @@
...
+##'
+##' aggregate_by_vector(x, k, colMeans)
+##' aggregate_by_vector(x, k, robustSummary)
+##' aggregate_by_vector(x, k, medianPolish)
+aggregate_by_vector <- function(x, INDEX, FUN, ...) {
+    if (!inherits(x, "matrix"))
+        stop("'x' must be a matrix.")
+    if (!identical(length(INDEX), nrow(x)))
+        stop("The length of 'INDEX' has to be identical to 'nrow(x).")
+    res <- by(x, INDEX, FUN, ...)
```
sgibb: Somehow this comment was deleted/not attached to my last review.
I would replace the `by` (that is just a wrapper around `tapply`) by a direct `tapply` call
to avoid the call of `as.matrix` and `as.list`:
```r
    res <- tapply(
        seq_len(nrow(x)),
        INDEX,
        FUN = function(i)FUN(x[i, , drop = FALSE], ...),
        simplify = FALSE
    )
    do.call(rbind, res)
```

R/aggregate.R
```diff
@@ -0,0 +1,175 @@
...
+    return(as.matrix(do.call(rbind, as.list(res))))
+}
+
+
+## aggregate_by_vector <- function(x, index, FUN, ...) {
```
sgibb:
```suggestion
## aggregate_by_list <- function(x, INDEX, FUN, ...) {    
```

R/aggregate.R
lgatto: Yes, that better indeed.

R/aggregate.R
lgatto: Done now

R/aggregate.R
lgatto: This would be an interesting QC for aggregation. I agree it would be better in an issue, but I feel that at this stage, it would be out of scope. Can we just leave it for now.

R/aggregate.R
lgatto: Done now.

R/aggregate.R
lgatto: Done now

R/aggregate.R
lgatto: Done now.

R/aggregate.R
lgatto: Done now.

R/aggregate.R
lgatto: Done now.

R/aggregate.R
lgatto: Done.

R/aggregate.R
lgatto: done now

R/aggregate.R
lgatto: Added Adriaan to the README.

R/aggregate.R
```diff
@@ -0,0 +1,180 @@
...
+    sample <- rep(colnames(x), each = nrow(x))[p]
+    feature <- rep(rownames(x), times = ncol(x))[p]
+
+    ## model.matrix breaks on factors with 1 level so make vector of
+    ## ones (intercept).
+    if (length(unique(sample)) == 1L) sample <- rep(1, length(sample))
+
+    ## Sum contrast on peptide level so sample effect will be mean
+    ## over all peptides instead of reference level.
+    X <- stats::model.matrix(~ -1 + sample + feature,
```
jorainer: Is there any reason to not import this function but to call it with `stats::`? Note that a call using `::` is slower than one without (according to Hadleys book - can't remember exactly where).

R/aggregate.R
```diff
@@ -0,0 +1,180 @@
...
+                             contrasts.arg = list(feature = 'contr.sum'))
+    ## MASS::rlm breaks on singulare values.
+    ## - Check with base lm if singular values are present.
+    ## - If so, these coefficients will be zero, remove this collumn
+    ##   from model matrix
+    ## - Rinse and repeat on reduced modelmatrix till no singular
+    ##   values are present
+    repeat {
+        fit <- stats::.lm.fit(X, expression)
```
jorainer: Also here, wouldn't it be better to import that function (see comment above).

R/aggregate.R
```diff
@@ -0,0 +1,180 @@
...
+        id <- fit$coefficients != 0
+        X <- X[ , id, drop = FALSE]
+        if (all(id)) break
+    }
+    ## Last step is always rlm: calculate estimated effects as
+    ## summarised values
+    fit <- MASS::rlm(X, expression, ...)
```
jorainer: Sorry for being so difficult today - to be correct we should have here a `requireNamespace` - which I would not really want to have, because this can slow things down. And we don't want to have `MASS` in `IMPORTS`?

R/aggregate.R
lgatto: No specific reason, just to avoid confusion - I wouldn't be surprised of there are other `model.matrix` functions around. 

In terms of speed, I honestly don't think it matters at all - I'm pretty sure this difference is insignificant in the wider context of aggregation.

R/aggregate.R
lgatto: Ok, will do that.

R/aggregate.R
lgatto: As above, given that it's the implementation of a fundamental statistical routine, I prefer explicitness over speed here. Not a hill I'm ready to die on, though.

R/aggregate.R
jorainer: Then I would add an `@importFrom stats model.matrix` to import it - you can then also be sure that that version/implementation is used.

sgibb:
We get a little bit confuse here. If `MASS` is not in `Imports` than we need the `::` operator and don't have to specify `importFrom(MASS, rlm)` in the `NAMESPACE`. That's why the current build fails.

We want to keep the dependencies as small as possible. That's why I want to set `MASS` into `Suggests`. @jorainer are you sure the minimal speed gain by avoiding `::` is worth an additional `Imports` dependency?

I don't have a strong argument here as `MASS` is a recommended package and should be available in nearly every R installation.

lgatto:
In terms of timings `stats::` is slightly slower but the difference is irrelevant.

```r
ff <- log(Volume) ~ log(Height) + log(Girth)
my.model.matrix <- model.matrix
> microbenchmark(model.matrix(ff, m), 
+                my.model.matrix(ff, m), 
+                stats::model.matrix(ff, m), 
+                times = 1000L)
Unit: microseconds
                       expr     min       lq     mean   median       uq     max
        model.matrix(ff, m) 151.967 154.5115 160.0858 156.5115 163.9150 322.487
     my.model.matrix(ff, m) 151.372 154.2000 160.0261 156.2455 163.8375 382.923
 stats::model.matrix(ff, m) 162.581 165.5750 171.4196 168.1185 175.9035 423.929
 neval
  1000
  1000
  1000
```
I will import in these 3 cases, given that both `MASS` and `stats` come with R but suggest we in other cases, we set for `Suggests` and `pkg::function` (possibly with a `requireNamespace("pkg")` for a proper error handling).

jorainer:
I usually find it cleaner to import the functions from the packages we need. Using `::` is OK, but IMHO needs a `requireNamespace` - and the combination of `requireNamespace` and `::` call can be slightly slower. For packages like `mzR` or some seldmonly used functions I'm OK with `::` and `requireNamespace`, but for most cases I would like to have the packages in `Imports` - but that's just my opinion of course.

lgatto:
Yes. This package is kind of special, as only a fraction of its functions will be used be each dependency at a time. Not sure what works best here.

Issue #30, opened 2019-12-11
Implement additional distance metrices


sgibb:
See also #29 

- Euclidean distance
- Absolute value distance

See: https://doi.org/10.1016/1044-0305(94)87009-8

@tnaake , @tobiasko any other needed?

tobiasko:
Normalized spectral angle

Maybe we should ask somebody from [ProteomeTools](http://www.proteometools.org/) what is used in practice?

michaelwitting:
What about having something similar to the functions used by GNPS?
https://ccms-ucsd.github.io/GNPSDocumentation/massspecbackground/networkingtheory/
Could be nice for the metabolomics community.

jorainer:
Nice idea @michaelwitting. Is there a publication, source code, reference implementation available?

michaelwitting:
I'm not aware of any implementation in R yet.

jmbadia:
We use **cosine similarity** as a measure to compare spectra in LC-MS/MS. Usually I apply the cosine() function from lsa package but now I have decided to use compareSpectra() with the following function
```
.weightxy <- function(x, y, m = 0, n = 0.5) {
    x ^ m * y ^ n
    }
cosSim <- function(x, y, m = 0L, n = 0.5, na.rm = TRUE) {
 wx <- .weightxy(x[, 1L], x[, 2L], m, n)
 wy <- .weightxy(y[, 1L], y[, 2L], m, n)
sum(wx^2L*wy^2L, na.rm = na.rm)/(sqrt(sum(wx^4L, na.rm = na.rm))*sqrt(sum(wy^4L, na.rm = na.rm)))
}
```
I read in your [pull](https://github.com/rformassspectrometry/MsCoreUtils/pull/33) that cosine similarity measure corresponds to the **standard** normalized dotproduct and that you are applying the other definition (**literature** definition). So I have modified the code you are using to implement the cosine similarity measure. It seems that works.

jorainer:
Great @jmbadia ! - feel free to make a pull request adding your function to `MsCoreUtils` (`"distance.R"` file) - maybe calling it `cosine` (or `ncosine` if it's using the weighting function @sgibb ?). If so, please include also unit tests comparing against results you would get from a reference implementation.

jmbadia:
Ok @jorainer. Happy to help.

jmbadia:
[...]

Issue #29, opened 2019-12-11
 Normalized spectral angle


sgibb:
distance measurement requested by @tobiasko

@tobiasko: could you please link an article with mathematical details here?

tobiasko:
Sure!

Toprak, U. H., Gillet, L. C., Maiolica, A., Navarro, P., Leitner, A., & Aebersold, R. (2014). Conserved peptide fragmentation as a benchmarking tool for mass spectrometers and a discriminating feature for targeted proteomics. Molecular & Cellular Proteomics : MCP, 13(8), 2056–2071. http://doi.org/10.1074/mcp.O113.036475

and maybe

Gessulat, S., Schmidt, T., Zolg, D. P., Samaras, P., Schnatbaum, K., Zerweck, J., et al. (2019). Prosit: proteome-wide prediction of peptide tandem mass spectra by deep learning. Nature Methods, 537(6), 347. http://doi.org/10.1038/s41592-019-0426-7

for an application and extension of normalized spectral angle (SA).

The topic of distance measures also touches spectrum transformation. (normalisation). Is there something planed/implemented? I see some connections to `scale(...)` in base R.

tobiasko:
Cold be nicely used for `.plotMirrorSpectra(...)`, but are you sure it belongs to `MSCoreUtils` @sgibb? Maybe `Spectra` is a better place?

sgibb:
Closed via d9cf17a938cbee9976e9c692b61d61fc32ea741d.

PR #33, opened 2019-12-11
Add some popular distance metrices

sgibb:
This PR implements a few distance/similarity measurements that are discussed in #29 and #30.

Would be great if @tobiasko could also have a look at the implementation.

The implemented scores are proposed in:
[Stein and Scott 1994](https://doi.org/10.1016/1044-0305(94)87009-8)
stein

[Toprak et al. 2014](https://doi.org/10.1074/mcp.O113.036475)
toprak

Please note that there is also a `.calibrate` function for scaling. I thought I would need it for the distance functions to scale them to `0:1` but it seems not to be necessary. But I would definitively need it for the `normalise` function. I can remove it if you like.


R/distance.R
```diff
@@ -0,0 +1,192 @@
...
+#'
+#' x <- matrix(c(1:5, 1:5), ncol = 2, dimnames = list(c(), c("mz", "intensity")))
+#' y <- matrix(c(1:5, 5:1), ncol = 2, dimnames = list(c(), c("mz", "intensity")))
+NULL
+
+#' @rdname distance
+#'
+#' @param m `numeric`, weighting for the first column of `x` and `y` (e.g.
+#' "mz"), default: `0` means don't weight by the first column.
+#' @param n `numeric`, weighting for the second column of `x` and `y` (e.g.
```
jorainer: very minor suggestion - could you eventually add a newline before every `@`? Documentation would get more readable in the source files (IMHO).

R/distance.R
```diff
@@ -0,0 +1,192 @@
...
+#'
+#' @examples
+#'
+#' x <- matrix(c(1:5, 1:5), ncol = 2, dimnames = list(c(), c("mz", "intensity")))
+#' y <- matrix(c(1:5, 5:1), ncol = 2, dimnames = list(c(), c("mz", "intensity")))
+NULL
+
+#' @rdname distance
+#'
+#' @param m `numeric`, weighting for the first column of `x` and `y` (e.g.
```
jorainer: Could you explain that parameter a little more? Is this a single numeric representing the weight for the complete first column or weights for each row? How is that used in the calculation?

R/distance.R
```diff
@@ -0,0 +1,192 @@
...
+#' "mz"), default: `0` means don't weight by the first column.
+#' @param n `numeric`, weighting for the second column of `x` and `y` (e.g.
+#' "intensity), default: `0.5` means effectly using `sqrt(x[,2])` and
+#' `sqrt(y[,2])`
+#' @param na.rm `logical(1)`, should `NA` removed prior to calculation (default
```
jorainer: `NA` be removed

tests/testthat/test_distance.R
```diff
@@ -0,0 +1,37 @@
+x <- matrix(c(1:5, 1:5), ncol = 2)
+y <- matrix(c(1:5, 5:1), ncol = 2)
+
+test_that("ndotproduct", {
```
jorainer: is there a way that we could cross-check with external results (i.e. results from another/original implementation of the methods)?

R/distance.R
sgibb: Of course.

R/distance.R
sgibb: Done.

R/distance.R
sgibb: It is specified in the details section of `ndotproduct` (and every other details section has a "please see details section of `ndotproduct`" line in it). Should I add a line "Please see `ndotproduct` details" or explain the arguments here (I am in favor of the "Please see" line)?

tests/testthat/test_distance.R
sgibb: That would be really great. But I only aware of our implementation in `MSnbase` and the one in `OrgMassSpecR`. The latter produce (of course) different results for the dotproduct. Especially the weighting isn't available in any tools I know:
https://github.com/OrgMassSpec/OrgMassSpecR/blob/aa0e4ee927b7f7051efa6400ad6f3fae8539f7c8/OrgMassSpecR/R/SpectrumSimilarity.R#L42
```r
similarity_score <- as.vector((u %*% v) / (sqrt(sum(u^2)) * sqrt(sum(v^2))))
```
vs our:
https://github.com/rformassspectrometry/MsCoreUtils/blob/b1e1ad095197d6ea558662348504202d8321a099/R/distance.R#L79-L84

Which (if `m = 0` and `n = 0.5`) boils down to:
```r
sum(sqrt(u) * sqrt(v))^2L / (sum(u) * sum(v))
# vs OrgMassSpecR
sum(u * v) / (sqrt(sum(u^2)) * sqrt(sum(v^2)))
```
As you see already this "simple" normalised dotproduct is hard to implement correctly.

@tobiasko, @tnaake are you aware of any software/web service were we can double check our results?

tobiasko: For the spectral contrast angle (SCA) I found this paper:

https://link.springer.com/content/pdf/10.1016/S1044-0305(01)00327-0.pdf

but they do simulations and list means/medians in the paper. But there is some basic geometric stuff in there that might help!

tobiasko: Hmmm...what about the following: Define s (actual position and intensity values do not matter), generate s' that is orthogonal in all dimension. Could be done by a function f(s) -> s'. The geometric definition of SCA would require that SCA of s vs. s` is 90 degree.

R/distance.R
jorainer: Yep I would do the same. Please add the *for more detail see*

tests/testthat/test_distance.R
sgibb: @tobiasko I am not sure that this is working because we the `dotproduct` is part of the `spectralangle` and we use a normalised dotpoduct here. So may we have to implement an unnoramlized dotproduct and spectral angle first?

R/distance.R
sgibb: Done

tests/testthat/test_distance.R
jorainer: I guessed already that it might be tricky. Suggestion: add a `@note` telling that the methods are implemented as described in xxx but that due to the lack of a reference implementation we can not check that the results are identical?

tobiasko: Is there already a function in `Spectra` that does the m/z binning? Using the def. from [Methods to Calculate Spectrum Similarity](https://link.springer.com/protocol/10.1007%2F978-1-4939-6740-7_7) one first derives a binned spectrum from the measured data, before calc. for instance the dot product:

"To create such vectors from acquired MS/MS spectra, each spectrum is divided into the same number of n bins, with n either set to a fixed-value [15, 22], or determined based on fragment ion tolerance [17]. Each bin is assigned a certain weight, calculated by summing up all peak intensities in that bin [22], or set to the highest peak intensity in that bin [17]. These binned spectra can now be considered as two vectors of equal dimensionality n, and can be matched against each other by the (normalized) dot product."

If I get this correctly, the binning makes sure that both spectra have aligned dimensions. How would one represent such a binned spectrum? The vectors will most likely be very sparse, at least at small bin sizes. Ideas?

tobiasko: @sgibb @jorainer Since there was no reply here is my version of a bin function. The function forces all m/z values of a spectrum (here a matrix) on a tuneable grid (default 1 u) and aggregates intensities by summation:

```
mzbin <- function(x, digits = 0){
  stopifnot(class(x) == "matrix", c("mz", "int") %in% dimnames(x)[[2]] )
  x[,"mz"] <- round(x[,"mz"], digits = digits)
  aggregate(int ~ mz, x, FUN = sum)
}
```

A basic demo:

```
> S1
        mz int
[1,] 100.0   1
[2,] 101.2   2
[3,] 101.4   3
> mzbin(S1)
   mz int
1 100   1
2 101   5
```
Simple, but pretty effective and it should work for big data sets (collections of Spectra)! What do you think?

tobiasko: Based on the above one can calc. a dot product for spectra in sparse matrix notation:
```
## dot product for vectors along m/z index (position)
dotProduct <- function(x, y){
  stopifnot(class(x) == "data.frame", class(y) == "data.frame",
            colnames(x) == c("mz", "int"), colnames(y) == c("mz", "int"))
  z <- merge(x, y, by = 'mz', all = TRUE) #full outer join
  sum(apply(z[,c("int.x", "int.y")], 1, prod), na.rm = TRUE) #sum of position-wise product
}
```
a small test:

```
> ## generate simple vectors along index 1, 2, ..., n
> v1 <- c(1, 2, 3)
> v2 <- c(4, 5, 6)
> v1 %*% v2 #dotp product along index
     [,1]
[1,]   32
> 
> ## create sparse spectrum along mz index in matrix notation
> S1 <- cbind(mz = c(100, 101.2, 102), int = v1)
> S2 <- cbind(mz = c(100, 100.91, 102), int = v2)
> v1 %*% v2 == dotProduct(mzbin(S1), mzbin(S2))
     [,1]
[1,] TRUE
```
:-)

tobiasko: The normalized dot product shoudl be:

```
> ## norm. dot productn for simple vectors
> x <- 1:3
> y <- 1:3
> 1 == (x %*% y)/(sqrt(sum(x^2)) * sqrt(sum(y^2)))
     [,1]
[1,] TRUE
```
same as `x%*%y / norm(x) * norm(y)`

tobiasko: and here is the norm. dot product along the lines above:

```
## norm. dot product for vectors along m/z dimension
normDotProduct <- function(x, y){
  stopifnot(class(x) == "data.frame", class(y) == "data.frame",
            colnames(x) == c("mz", "int"), colnames(y) == c("mz", "int"))
  z <- merge(x, y, by = 'mz', all = TRUE) #full outer join
  dp <- sum(apply(z[,c("int.x", "int.y")], 1, prod), na.rm = TRUE) #sum of position-wise product
  normx <- sqrt(sum(x[,"int"]^2))
  normy <- sqrt(sum(y[,"int"]^2))
  dp / (normx * normy)
}
```
test for edge cases:

```
> ## test
> ## case 1: all peaks match on 1 u grid 
> S1 <- cbind(mz = c(100, 101.2, 102), int = 1:3)
> S2 <- cbind(mz = c(100, 100.91, 102), int = 1:3)
> 1 == normDotProduct(mzbin(S1), mzbin(S2))
[1] TRUE
> 
> ## case 2 : no peaks match on 1 u grid
> S2 <- cbind(mz = c(200, 200.91, 202), int = 1:3)
> 0 == normDotProduct(mzbin(S1), mzbin(S2))
[1] TRUE
```
This is far more powerful than [R/dotproduct.R](https://github.com/rformassspectrometry/MsCoreUtils/blob/master/R/dotproduct.R)
which just warns if peaks are not aligned and still computes a value.

tobiasko: A more general question @sgibb : Why do you want to place these functions outside of the `Spectra` package (MSCoreUtils)? Why not simply in `Spectra`? A "." version for internal usage and versions that take `Spectra` objects as arguments (exported) and are fully vectorised?

sgibb: @tobiasko: thanks for your code suggestions but we already have a [binning](https://rformassspectrometry.github.io/MsCoreUtils/reference/binning.html) and an outer [join](https://rformassspectrometry.github.io/MsCoreUtils/reference/matching.html).

`dotproduct` etc. are needed in `Chromatograms` as well. That's why we put it here. We already have a `compareSpectra` in `Spectra`.

> This is far more powerful than R/dotproduct.R

Why? And `R/dotproduct.R` is obsolete with this PR.

tobiasko: Why? Because...
```
> ## copied from
> ## https://github.com/rformassspectrometry/MsCoreUtils/blob/distance/R/distance.R
> 
> .weightxy <- function(x, y, m = 0, n = 0.5) {
+   x ^ m * y ^ n
+ }
> 
> ndotproduct <- function(x, y, m = 0L, n = 0.5, na.rm = TRUE) {
+   wx <- .weightxy(x[, 1L], x[, 2L], m, n)
+   wy <- .weightxy(y[, 1L], y[, 2L], m, n)
+   sum(wx * wy, na.rm = na.rm)^2L /
+     (sum(wx^2L, na.rm = na.rm) * sum(wy^2L, na.rm = na.rm))
+ }
> 
> ## test on case with no shared peaks
> x <- matrix(c(6:10, 1:5), ncol = 2, dimnames = list(c(), c("mz", "intensity")))
> y <- matrix(c(1:5, 1:5), ncol = 2, dimnames = list(c(), c("mz", "intensity")))
> ndotproduct(x, y)
[1] 1
> # :-(
```
and that should be zero by definition!? Please compare to case 2 above.

sgibb: These is expected. As you mentioned above all distance/similarity measurements require binned spectra. The input in your example is not aligned but `ndotproduct` assumes it is (and for that an intensity of `1:5` against an intensity `1:5` should return a similarity of `1`). So a fair comparison would be:
```r
joinndotproduct <- function(x, y) {
    j <- join(x[, "mz"], y[, "mz"], type = "outer")
    ndotproduct(x[j$x,], y[j$y,])
}

joinndotproduct(x, y)
# [1] 0
```
That is exactly what `Spectra::compareSpectra` is doing. `compareSpectra` is more powerful because it allows the user to select a joining function (outer/inner/left/right/(graph based if the other PR will be accepted)/user provided) and a distance/similarity function (one of these that are part of this PR or any other function the user provides).

We try to build functions for just one task (following the unix philosophy: doing on thing and do it well. Your `normDotProduct` does at least 2 task: joining and distance measurement).

sgibb: As already mentioned above my problem with the dotproduct is the following:

The general definition of a normalized dotproduct is:
```
sum(u * v) / (sqrt(sum(u^2)) * sqrt(sum(v^2))
```
but in the literature (e.g. Stein and Scott 1994) it is 
```
sum(sqrt(u) * sqrt(v))^2L / (sum(u) * sum(v)))
```
(for the default weighting with m = 0, and n = 0.5)

This will definitively yield different results.

jorainer: Also providing some background info: we did separate the binning/peak matching and similarity calculation steps because we wanted to give the users the choice to select a specific method for each of these steps (i.e. some people might prefer binning the data while others might want to match peaks based on difference of their m/z). The similarity calculations are performed on the matched (or *aligned* spectra). It would thus also be possible to use different, base R, similarity calculation functions.

For the reason to have these in `MsCoreUtils` and not in `Spectra`: some/many of these functions will be re-used in the `Chromatograms` package.

sgibb:
@jorainer I added the note you suggested. Could we merge this now?

sgibb:
BTW: the travis failure seems to be related to a `S4Vectors` problem upstream.

jorainer:
OK for me @sgibb - you can merge.

jorainer:
Regarding `S4Vectors`, strangely enough travis for PR #45 does not fail...

lgatto:
I apologise in advance, because this will lead to a *snake_case* vs *camelCase* discussion... 

I initially found the leading *n* confusing, and then inferred from the documentation that these returned normalised distances. What about a more telling name, starting with *norm* or *normali[z|s]ed*.

jorainer:
Honestly - I prefer `dotproduct`/`ndotproduct` over `dotproduct`/`normDotproduct` (or `norm_dotproduct` or `normdotproduct`)... but I agree, this should be clearly (better?) described in the documentation.

sgibb:
I don't really like `normDotProduct` etc. (neither `norm_dot_product`) and I am happy to extend the documentation. Every section starts with 
> `ndotproduct`: the normalized dot product ...

Any suggestions for improvement?

jorainer:
maybe one general section where you specify that function names for methods that produce *normalized* similarities start with an *n*?

sgibb:
Ok, I a few more words.

sgibb:
Merged manually due to conflicts introduced by the addition of imputation/normalisation and bioc submission.

PR #40, opened 2020-01-30
Adds imputation functions 

lgatto:
This PR adds the main `impute_matrix` function and a set of helper functions that perform imputation, their documentation (all in `imputation.Rd`) and unit tests (in `test_imputation.R`).

Once this is merged, these functions will be removed from and used in `Features`.


lgatto:
Small note to make the reviews easier - this code is mostly ported from `MSnbase` via `Features`.

R/imputation.R
```diff
@@ -0,0 +1,326 @@
...
+##' 10.1021/acs.jproteome.5b00981. PubMed PMID:26906401.
+##'
+##' @rdname imputation
+##'
+##' @aliases imputeMethods impute_neighbour_average impute_knn impute_mle impute_bpca impute_mixed impute_min impute_zero impute_matrix
+##'
+##' @importFrom Rcpp sourceCpp
+##' 
+##' @useDynLib MsCoreUtils
+##'
```
jorainer: I would also add an `@author Laurent Gatto` here - so that we know who to blame if something is nor working ;)

R/imputation.R
```diff
@@ -0,0 +1,326 @@
+##' @title Quantitative mass spectrometry data imputation
```
jorainer: very minor point: so far we used `#'` for roxygen comments - just for consistency reasons I think it might be better to replace these `##'` with `#'`.

R/imputation.R
```diff
@@ -0,0 +1,326 @@
...
+##'
+##' MNAR features should ideally be imputed with a left-censor method,
+##' such as `QRILC` below. Conversely, it is recommended to use host
+##' deck methods such nearest neighbours, Bayesian missing value
+##' imputation or maximum likelihood methods when values are missing
+##' at random.
+##'
+##' Currently, the following imputation methods are available. 
+##'
+##' - *MLE*: Maximum likelihood-based imputation method using the EM
```
jorainer: would it be more intuitive for the user if you used here `method = "MLE"` instead of *MLE*?

R/imputation.R
```diff
@@ -0,0 +1,326 @@
...
+    c("bpca","knn", "QRILC", "MLE",
+      "MinDet", "MinProb", "min", "zero",
+      "mixed", "nbavg", "none")
+
+##' @export
+##' @rdname imputation
+##' @param k `numeric(1)` providing the imputation value used for the
+##'     first and last samples if they contain an `NA`. The default is
+##'     to use the smallest value in the data.
+impute_neighbour_average <- function(x, k = NULL) {
```
jorainer: oh yes! it's snake_case time! wondering what @sgibb will say to this :) I personally like it this way, especially since function names like `imputeKnn` would look strange...

R/imputation.R
```diff
@@ -0,0 +1,326 @@
...
+        stop("Number of proteins and length of randna must be equal.",
+             call. = FALSE)
+    x[randna, ] <- impute_matrix(x[randna, ], mar, ...)
+    x[!randna, ] <- impute_matrix(x[!randna, ], mnar, ...)
+    x
+}
+
+##' @export
+##' @rdname imputation
+impute_min <- function(x) {
```
jorainer: Would we also want to allow users to replace `NA` also with only a fraction/proportion of the min? Something like

```r
impute_min <- function(x, prop = 1) {
    val <- min(x, na.rm = TRUE) * prop
    x[is.na(x)] <- val
    x
}
```

so it would also be possible to replace `NA`s with half the minimum `impute_min(x, prop = 0.5)`.

R/imputation.R
lgatto: :-)

R/imputation.R
lgatto: But I would need to repeat `method = ` every time. I think, given that method is mentioned before the list, it should be fine.

R/imputation.R
lgatto: I am not sure about that. I would then suggest a `impute_with`, and users can set to whatever they want, for example `min(x, na.rm = TRUE) * 0.5`. I will also add that one as an example.

R/imputation.R
jorainer: OK

PR #41, opened 2020-01-31
Add normalize_matrix function

lgatto:


tests/testthat/test_normalize.R
```diff
@@ -0,0 +1,6 @@
+test_that("function: all normalize methods", {
+    for (.method in normalizeMethods()) {
```
jorainer: Are you planning to add some more unit tests here, e.g. to check that the methods actually do what they should?

R/normalize.R
```diff
@@ -0,0 +1,98 @@
...
+##' `"max"`, `"quantiles"`, `"center.mean"`, `"center.median"`,
+##' `"center.median"`, `"quantiles.robust`" or `"vsn"`.  For `"sum"`
+##' and `"max"`, each feature's intensity is divided by the maximum or
+##' the sum of the feature respectively. These two methods are applied
+##' along the features (rows). The `normalizeMethods()` function
+##' returns a vector of available normalisation methods.
+##'
+##' `"center.mean"` and `"center.median"` translate the respective
+##' sample (column) intensities according to the column mean or
+##' median. `"diff.median"` translates all samples (columns) so that
```
jorainer: I think it would be good if you mentioned here that in fact the column mean or median is **subtracted** from each value of that column. I usually normalize the data by dividing the column intensities by the column median / median across all column medians - so, applying this method I was first surprised by its result.

Also, I was at first puzzled by the word *translate* here - maybe something like *centers* (or *shifts*) might be better?

jorainer:
Nice PR @lgatto ! I would just provide some more details about how the actual calculations are preformed in the documentation - that might be helpful since these functions are tailored more for the developers.

tests/testthat/test_normalize.R
lgatto: This could be done, indeed. I am not sure to what extend is will be easy to do though, in some cases at least. Such unit tests should probably be added upstreams... 
But for now, I wanted to make sure all functions could be called.

R/normalize.R
lgatto: Thank you, that's really misleading. I will change this and add the versions that divide in the next PR.

tests/testthat/test_normalize.R
jorainer: sure, totally understand - also, by looking at the code it is pretty straight forward and one might expect the functions to work properly...

R/normalize.R
jorainer: Regarding division *vs* subtraction, I'm totally fine with the subtraction - if one wanted to use a division instead it would be pretty straight forward by first log2 transforming the data. So, the subtraction is OK, I would just clarify this in the documentation.

lgatto: I think the best is the have `center.mean` and `center.media` using the subtraction (that's how `scale` works) and add `div.mean` and `div.median`. What do you think?

lgatto: Note that it was possible to set your expected behaviour by using `normalize_matrix(x, "center.mean", FUN = "/")`, but that wasn't clear, so I prefer to add the `div.*` methods.

jorainer: Sure, makes sense.

lgatto:
@jorainer - I think this last PR addresses all your points.

PR #45, opened 2020-02-11
feat: add rla and rowRla functions

jorainer:
- Add functions to calculate *relative log abundances*.
- Add documentation and related unit tests.


jorainer:
These *RLA* values are very helpful for quality assessment. I always use them to evaluate normalization performance. The reference is [this](https://pubs.acs.org/doi/10.1021/ac302748b) - it's essentially the same as the RLE values on Affymetrix gene chips - for those that still remember this kind of data :)

R/rla.R
```diff
@@ -0,0 +1,80 @@
...
+#'
+#' rla(x, grp)
+#'
+#' x <- rbind(c(324, 4542, 3422, 3232, 5432, 6535, 3321, 1121),
+#'      c(12, 3341, 3034, 6540, 34, 4532, 56, 1221))
+#' grp <- c("a", "b", "b", "b", "a", "b", "a", "b")
+#'
+#' ## row-wise RLA values
+#' rowRla(x, grp)
+rla <- function(x, f, log.transform = TRUE, na.rm = TRUE) {
```
sgibb: I would use a default argument for `f` instead of the following `if (missing(f))` statement.
```suggestion
rla <- function(x, f = rep_len(1, length(x)), log.transform = TRUE, na.rm = TRUE) {
```

R/rla.R
```diff
@@ -0,0 +1,80 @@
...
+    if (missing(f))
+        f <- rep_len(1, length(x))
+    if (length(x) != length(f))
```
sgibb: This is handled by `split.default` (there a warning is thrown). Do we want to convert it into an error?

jorainer: But there is no `warning` if `x` is a multiple of `x`. I personally don't like that behaviour, that's why I would check it here.

jorainer:
I like the idea of changing the `log.transform`. But then I would use `transform = c("log2", "log10", "none")` instead.

The `md` (median deviation) - I had to look several times at the code, but it seems that your `md` is not calculating a median deviation for each group, but *simply* subtracting the median of each group from each value. For `md` I would expect to get a `numeric` of length equal to the number of groups and not a `numeric` of length equal `x`.

R/rla.R
sgibb: Ok, fine.

sgibb:
For the `md` I just copied your `rla` code and shortened it. I not even tested it. It was just for illustration what my idea would be. For the "tidy"/pipe syntax it would be something like: `x %>% log %>% md`

jorainer:
Changed the parameter `log.transform` to `transform` as discussed above. I don't see a big benefit for generalizing this function (with e.g. the `md`), but if so, I would call it `median_difference` instead of `md`. End users might not understand that with a `x %>% log2 %>% median_difference` they would get an RLA.

R/rla.R
```diff
@@ -0,0 +1,82 @@
...
+#' x <- rbind(c(324, 4542, 3422, 3232, 5432, 6535, 3321, 1121),
+#'      c(12, 3341, 3034, 6540, 34, 4532, 56, 1221))
+#' grp <- c("a", "b", "b", "b", "a", "b", "a", "b")
+#'
+#' ## row-wise RLA values
+#' rowRla(x, grp)
+rla <- function(x, f = rep_len(1, length(x)),
+                transform = c("log2", "log10", "identity"), na.rm = TRUE) {
+    transform <- match.arg(transform)
+    if (missing(f))
```
sgibb: This `if` statement shouldn't be necessary anymore.

R/rla.R
```diff
@@ -0,0 +1,82 @@
...
+    grp_meds <- unlist(lapply(split(x, f), median, na.rm = na.rm))
+    res <- x - grp_meds[f]
+    names(res) <- names(x)
+    res
+}
+
+#' @rdname rla
+#'
+#' @export
+rowRla <- function(x, f, transform = c("log2", "log10", "identity")) {
```
sgibb: I guess we have to specify the default `f` here as well to remove the `if(missing(f))` above:
```suggestion
rowRla <- function(x, f = rep_len(1, ncol(x)), transform = c("log2", "log10", "identity")) {
```

R/rla.R
jorainer: yep, you're right.

jorainer: removed now.

PR #52, opened 2020-05-04
Adding functions for spectra mining

michaelwitting:


sgibb:
Dear Michael,

thanks for your contribution but I am wondering whether we really need these functions. 
They are more or less trivial one-liners:

```r
any(common(precursorMz - neutralLoss, x[, "mz"], tolerance = 0.005))
any(common(productMz, x[, "mz"], tolerance = 0.005))
```

Do you use them that often to justify the additional code in `MsCoreUtils`?

michaelwitting:
Hi Sebastian,
I can only speak for myself. I'm using this function quite often to check for different structural motifs in several thousands of fragmentation spectra from lipidomics data. It seems clear to me that these one liners can do the same thing. However, coming from the wet-lab side and speaking for some other wet-lab people it might feel better and more appropriate to have distinct functions.

jorainer:
I see @sgibb's point. The `MsCoreUtils` package is intended to be used mostly by developers and should thus contain only a reduced set of functions (sorry, was my fault to suggest to add these functions). Unless the functions involve also somehow the intensity values I would also not add them as dedicated functions in `MsCoreUtils` **but** I think it would help to add this functionality to `Spectra`:

Both, the neutral loss and product ion functions check if a certain `mz` value is present. What if we added a function `hasMz` or `containsMz` function for `Spectra` objects that checks if a specified m/z value is present in each spectrum? The help page/examples can then list as possible use cases the neutral loss and product ion search.

michaelwitting:
I'm fine with both ways. I was thinking of having another argument called `relIntensityTreshold` to filter out m/z values below a certain relative treshold and then perform the checks. This way you can make sure that you are not searching in the noise.
Otherwise, it can go to `Spectra`. Fine with the name `hasMz` or `containsMz`, just for a wet-lab person that might want to use that it is not completly obvious what they can do with it. We have to be very exact then in the help ;-)

sgibb:
> `containsMz` function for Spectra objects that checks if a specified m/z value is present in each spectrum? The help page/examples can then list as possible use cases the neutral loss and product ion search.

:+1: for `containsMz` in `Spectra` as general function. Maybe we could add two one line functions for the most common use cases (product ion, neutral loss) as well.

PR #53, opened 2020-05-22
fix: deprecate doproduct signature

lgatto:


lgatto:
Once the PR is accepted and merged, I will push to Bioc devel.

PR #54, opened 2020-05-27
feat: new countFeatures aggregation function

lgatto:


R/aggregate.R
```diff
@@ -108,6 +110,33 @@ medianPolish <- function(x, verbose = FALSE, ...) {
     medpol$overall + medpol$col
 }
 
+##' @title Counts the number of features
+##'
+##' @description
+##' Returns the number of non-NA features if a features by sample
```
jorainer, line 116: small typo, guess it should read *in a features by sample*

R/aggregate.R
```diff
@@ -108,6 +110,33 @@ medianPolish <- function(x, verbose = FALSE, ...) {
...
+##' 
+##' @author Laurent Gatto
+##'
+##' @examples
+##' m <- matrix(c(1, NA, 2, 3, NA, NA, 4, 5, 6),
+##'             nrow = 3)
+##' countFeatures(m)
+##' m <- matrix(rnorm(30), nrow = 3)
+##' countFeatures(m)
+countFeatures <- function(x, ...)
```
jorainer, line 137: I was wondering, because the `MsCoreUtils` is supposed to contain plain *generic* functions and the term *feature* is already quite specific and can have different meanings in different domains, would it not be better to call the function `colCounts`? This would be somehow along the `colSums`, `colMeans` etc functions, but this one would simply count the number of non-NA values per column.

lgatto:
Requested changes implemented. I also added some more unit tests.

lgatto:
Thank you @jorainer - all your requests have been addressed!

PR #58, opened 2020-06-23
Fix rbindFill to work also on a single data.frame

jorainer:
`rbindFill` failed if a single `data.frame` was passed to the function. This PR fixes this.


sgibb:
There are some git conflicts in this PR, so I close this in favour of #59

PR #59, opened 2020-06-23
Fix rbindFill single dframe bug

sgibb:
Follow-up to #58


jorainer:
Are you pushing this to BioC @sgibb ?

PR #62, opened 2020-08-07
fix: ensure closest accept just tolerance of length 1 or ntable; closes #61

### LINKED ISSUE #61: Potential bug in closest?
@sgibb we have `tolerance <- rep_len(tolerance, ntable) + ppm(table, ppm) +` in the `closest` function (see [here](https://github.com/rformassspectrometry/MsCoreUtils/blob/cfbbb0c6de25eeb8a258a4600888bb9135283311/R/matching.R#L109)) but as far as I remember the idea was that we would support `tolerance` to be of length 1 or equal `length(table)` but then this line should be `tolerance <- tolerance + ppm(table, ppm) + ...` instead. Can you please cross-check?

sgibb:
This PR introduces a new `if` statement to catch `tolerance` that arn't 1 or `ntable`.


sgibb:
Indeed! Thanks!

PR #63, opened 2020-08-07
Implement peak joining functions in C

jorainer:
While the `join*` functions are very elegantly implemented by @sgibb in R, the matching between peaks is still the biggest bottleneck in spectrum comparison.

This PR adds an implementation of the outer join in C:

```r
library(MsCoreUtils)
library(microbenchmark)
x <- as.numeric(c(1, 2, 5, 6))
y <- as.numeric(2:5)
microbenchmark(
    MsCoreUtils:::.joinOuter(x, y, 0, 0),
    MsCoreUtils:::.cjoinOuter(x, y, 0, 0)
)
Unit: microseconds
                                  expr   min     lq    mean median     uq   max
  MsCoreUtils:::.joinOuter(x, y, 0, 0) 108.5 110.65 127.316 111.95 119.55 221.2
 MsCoreUtils:::.cjoinOuter(x, y, 0, 0)  17.7  18.50  21.936  20.10  21.45 103.5
 neval cld
   100   b
   100  a 
```

I did also a lot of comparison (also on a data set of ~ 500,000 real spectra) to check that both functions yield the same result. They are always identical with the exception that the original implementation has a problem with ties, i.e. if in one of the vectors values are duplicated:

```r
a <- c(1, 2, 4, 4, 5)
b <- c(2, 3, 4, 5, 6)
MsCoreUtils:::.joinOuter(a, b, 0, 0)
$x
[1]  1  2 NA NA  4  5 NA

$y
[1] NA  1  2 NA  3  4  5

MsCoreUtils:::.cjoinOuter(a, b, 0, 0)
$x
[1]  1  2 NA  3  4  5 NA

$y
[1] NA  1  2  3 NA  4  5

```

The original implementation drops/skips element `3` from `a` entirely while the new implementation reports the first mapping.

Happy to discuss if this is worth the effort - and even more happy if someone can come up with a more elegant (and faster) implementation in C!

If we agree I would also implement the other join operations in C and change `join` to use the C functions.

PR #64, opened 2020-08-14
refactor: implement closest in C

sgibb:
This is a follow-up PR to #63.

Changes:

1. Reimplementation of `closest` in C.
2. Introduction of a new argument `.check`.

---

## Reimplementation of `closest` in C

This PR introduces 5 new C functions:
- `leftmost`.
- `C_closest_dup_keep`.
- `C_closest_dup_closest`.
- `closest_dup_closest`.
- `C_closest_dup_remove`.

The workhorse function is `leftmost`. It is very similar to R's internal
`findInterval` but simplified for our use cases.
It finds the leftmost value of `x` in `table` via an *elastic search*.
The for `*closest_dup*` functions just loop through `x`, test the return value
of `leftmost` against some constraints (e.g. `tolerance`) and handle duplicated
matches in different ways. The reason for 3 `C_closest_dup*` functions is that
I don't want to have an additional `if` statement inside the loop to reduce
runtime (as described in #63 the `join` functions that are based on `closest`
are applied > million times). So I think that in this special case code
duplication is worth the effort.

The `closest_dup_closest` function contains the body of the API function
`C_closest_dup_closest` because I am going to use it in the C implementation of
`joinOuter`.

In contrast to the former implementation using `findInterval` the current
strictly expects non-decreasingly sorted values for `x` and `table` without
`NA`. Before `x` could be unsorted and contain `NA`. I don't think that our
use cases need `NA`. In general mass spectrometry data should be sorted as well.
So I assume no problems here. Handling unsorted data for `x` would blow up the
code with more `if`/`else` statements and may increase the runtime.

Please note that the current implementation uses `R_xlen_t` instead of `int`
for indices and `REALSXP` instead of `INTSXP` for the output value to support
long vectors. Long vectors are not supported for `findInterval`.

This new code offers the opportunity to change the way the `tolerance`/`ppm`
arguments are used. I can't find the issue but if I remember correctly @jorainer
was astonished by the fact that `tolerance` has to have the length of `table`
and not of `x`. That was due to the way `closest` was vectorized in R. Now we
could choose to have `length(tolerance) == length(table)` (as in the original R
implementation and in the current PR) or `length(tolerance) == length(x)`.
Any wishes?

## Introduction of a new argument `.check`

This PR additionally introduces a new `.check` argument to `closest`, `common`,
 and `join`. The default is `.check = TRUE` which will run `is.unsorted` on `x`
and `table`. Running `is.unsorted` twice takes often equal/more time as/than the
whole elastic search (depending on the length and distribution of `x` and
`table`). Because `Spectra` guarantees that mz values are sorted I like to
provide this argument to disable these checks via `.check = FALSE` for maximum
speed. The average user should not use this argument at all.
An alternative would be to provide an `.closest`, `.common`, etc. which
don't use `is.unsorted` (and keep the user-visible `closest` with `is.unsorted`
as it is done now; but I don't like the R CMD check warning: ":::" was used).

## Benchmarks

``` r
library("MsCoreUtils")
#>
#> Attaching package: 'MsCoreUtils'
#>
#>     smooth

closest.R <- function(x, table, tolerance = Inf, ppm = 0,
                    duplicates = c("keep", "closest", "remove"),
                    nomatch = NA_integer_) {

    if (!all(is.numeric(tolerance)) || any(tolerance < 0))
        stop("'tolerance' has to be a 'numeric' larger or equal zero.")

    if(!all(is.numeric(ppm)) || any(ppm < 0))
        stop("'ppm' has to be a 'numeric' larger or equal zero.")

    if (!is.numeric(nomatch) || length(nomatch) != 1L)
        stop("'nomatch' has to be a 'numeric' of length one.")

    table <- table[!is.na(table)]
    ntable <- length(table)
    if (!ntable)
        return(rep_len(nomatch, length(x)))

    tolerance <- rep_len(tolerance, ntable) + ppm(table, ppm) +
        sqrt(.Machine$double.eps)
    duplicates <- match.arg(duplicates)

    lIdx <- findInterval(x, table, rightmost.closed = FALSE, all.inside = TRUE)
    rIdx <- lIdx + 1L

    lIdx[lIdx == 0L] <- 1L
    rIdx[rIdx > ntable] <- ntable

    lDiff <- abs(table[lIdx] - x)
    rDiff <- abs(table[rIdx] - x)

    lDiff[lDiff > tolerance[lIdx]] <- Inf
    rDiff[rDiff > tolerance[rIdx]] <- Inf

    d <- which(lDiff > rDiff)

    lIdx[d] <- rIdx[d]

    ## no match at all
    lIdx[is.infinite(lDiff) & is.infinite(rDiff)] <- NA_integer_

    ## duplicated matches
    if (duplicates == "remove") {
        lIdx[duplicated(lIdx) | duplicated(lIdx, fromLast = TRUE)] <-
            NA_integer_
        lIdx[is.finite(lDiff) & is.finite(rDiff)] <- NA_integer_
    } else if (duplicates == "closest") {
        ## lIdx could be updated
        o <- order(abs(table[lIdx] - x))
        m <- lIdx[o]
        m[duplicated(m)] <- NA_integer_
        lIdx[o] <- m
    }

    if (!identical(nomatch,
 NA_integer_))
        lIdx[is.na(lIdx)] <- nomatch

    as.integer(lIdx)
}

set.seed(42)
x <- sort(sample(1000, 100) + runif(100))
y <- sort(sample(1000, 700))
```

### "keep"

```r
library("microbenchmark")
microbenchmark(
    closest(x, y),
    closest(x, y, .check = FALSE),
    closest.R(x, y),
    check = "equal"
)
#> Unit: microseconds
#>                           expr    min      lq     mean  median      uq     max
#>                  closest(x, y) 39.191 39.9835 42.29627 40.9110 42.3475 121.462
#>  closest(x, y, .check = FALSE) 34.188 34.7645 36.37649 35.7070 36.7445  51.758
#>                closest.R(x, y) 67.566 69.7350 73.20936 71.8575 74.9485 113.451
#>  neval cld
#>    100  b
#>    100 a
#>    100   c
microbenchmark(
    closest(y, x),
    closest(y, x, .check = FALSE),
    closest.R(y, x),
    check = "equal"
)
#> Unit: microseconds
#>                           expr     min       lq      mean   median       uq
#>                  closest(y, x)  41.036  42.3145  44.17169  43.4235  45.2095
#>  closest(y, x, .check = FALSE)  35.965  37.0555  39.62806  37.7740  40.1550
#>                closest.R(y, x) 104.209 106.2965 109.07885 107.5940 109.7015
#>      max neval cld
#>   59.948   100  b
#>  123.258   100 a
#>  156.641   100   c
```

### "closest"

```r
microbenchmark(
    closest(x, y, duplicates = "closest"),
    closest(x, y, duplicates = "closest", .check = FALSE),
    closest.R(x, y, duplicates = "closest"),
    check = "equal"
)
#> Unit: microseconds
#>                                                   expr     min      lq
#>                  closest(x, y, duplicates = "closest")  43.989  45.910
#>  closest(x, y, duplicates = "closest", .check = FALSE)  38.720  39.916
#>                closest.R(x, y, duplicates = "closest") 134.073 140.269
#>       mean   median       uq     max neval cld
#>   50.44183  48.5235  53.7565  78.640   100  a
#>   46.15176  43.0545  48.2810 179.536   100  a
#>  147.89830 143.4540 148.1075 266.834   100   b
microbenchmark(
    closest(y, x, duplicates = "closest"),
    closest(y, x, duplicates = "closest", .check = FALSE),
    closest.R(y, x, duplicates = "closest"),
    check = "equal"
)
#> Unit: microseconds
#>                                                   expr     min       lq
#>                  closest(y, x, duplicates = "closest")  49.733  52.6950
#>  closest(y, x, duplicates = "closest", .check = FALSE)  44.118  46.3405
#>                closest.R(y, x, duplicates = "closest") 242.353 245.6805
#>       mean   median      uq     max neval cld
#>   57.76719  56.0315  62.855  74.344   100  b
#>   51.41085  48.9930  56.263  88.473   100 a
#>  252.69544 247.8425 251.732 467.287   100   c
```

<sup>Created on 2020-08-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

EDIT: somehow my `hub` command send the PR four days ago while I was preparing the PR message, sorry. Now its finalized and ready for review.


jorainer:
Excellent! I totally agree with all the points you mentioned above (i.e. code duplication instead of `if/else` in the C code, avoiding to call `is.unsorted` unnecessarily, ...). Regarding the `tolerance`, I would change it to have `length(x)` instead of `length(table)` - imagine if you're trying to match one spectrum against a library/database. Then you know the number of peaks you have in your query spectrum but don't know that of the target spectra. Thus, to me, supporting `tolerance` to be either of length 1 or `length(x)` is more intuitive.

R/matching.R
```diff
@@ -101,7 +115,13 @@ closest <- function(x, table, tolerance = Inf, ppm = 0,
     if (!is.numeric(nomatch) || length(nomatch) != 1L)
         stop("'nomatch' has to be a 'numeric' of length one.")
 
-    table <- table[!is.na(table)]
+    if (.check && (
+            !identical(FALSE, is.unsorted(x) ||
```
jorainer, line 119: why `identical` here and not simply `.check && (is.unsorted(x) || is.unsorted(table))` ?

R/matching.R
```diff
@@ -112,44 +132,30 @@ closest <- function(x, table, tolerance = Inf, ppm = 0,
         stop("'tolerance' hat to be of length 1 or equal to 'length(table)'")
```
jorainer, line 132: could we change this now to have length equal `length(x)`?

R/matching.R
```diff
@@ -112,44 +132,30 @@ closest <- function(x, table, tolerance = Inf, ppm = 0,
...
-        m[duplicated(m)] <- NA_integer_
-        lIdx[o] <- m
-    }
-
-    if (!identical(nomatch, NA_integer_))
-        lIdx[is.na(lIdx)] <- nomatch
+    duplicates <- match.arg(duplicates)
 
-    as.integer(lIdx)
+    if (duplicates == "keep")
```
jorainer, line 138: what if you used `.Call(paste0("C_closest_dup_", duplicates), ...)`? We could get rid of all these `if` `else` statements - and maybe it would be even slightly faster?

src/closest.c
```diff
@@ -0,0 +1,235 @@
...
+SEXP C_closest_dup_keep(SEXP x, SEXP table, SEXP tolerance, SEXP nomatch) {
+    R_xlen_t nx = XLENGTH(x);
+    double* px = REAL(x);
+
+    R_xlen_t ntable = XLENGTH(table);
+    double* ptable = REAL(table);
+
+    double* ptolerance = REAL(tolerance);
+
+    SEXP out = PROTECT(allocVector(REALSXP, nx));
```
jorainer, line 81: I did not quite undestand the benefit of using here `REALSXP` instead of `INTSXP` - I would have expected the index to be returned as an integer not as a real.

jorainer (changes requested):
Excellent @sgibb ! I really like this PR! Two things however: 
- I think it might be more user friendly to allow `tolerance` to be of length equal `length(x)`.
- I don't (yet) see the benefit of returning a `numeric` instead of an `integer` from the function. Could you please explain again?

Also, how are we going to deal with `NA` values? What would happen if we had an `NA` in either `x` or `table`?

R/matching.R
sgibb: If `x` or `table` contain `NA`, `is.unsorted` return `NA` as well, e.g. `is.unsorted(c(1, NA))`. So I have to catch both cases which could easily be done with `identical`.

R/matching.R
sgibb: Of course. As I said this is now easily possible but will be a backward-incompatible change and may already break something (for `Spectra`).

R/matching.R
sgibb: My initial idea was that the R compiler could better work with an `if`/`else` than with a `paste` but that seems not to be the case. Nevertheless `paste0` is slower:
```r
devtools::load_all()
library("compiler")

set.seed(42)
x <- sort(sample(1000, 100) + runif(100))
y <- sort(sample(1000, 700))

cmpIfElse <- cmpfun(cIfElse)
cmpPaste <- cmpfun(cPaste)

library("microbenchmark")
microbenchmark(cIfElse(x, y), cPaste(x, y), cmpIfElse(x, y), cmpPaste(x, y))
# Unit: microseconds
#             expr    min      lq     mean  median      uq     max neval
#    cIfElse(x, y) 25.799 26.3710 27.54267 26.7995 28.5360  41.525   100
#     cPaste(x, y) 29.878 30.4775 31.42049 30.8500 32.6400  34.705   100
#  cmpIfElse(x, y) 24.651 25.6600 28.64918 26.0425 27.7225 217.980   100
#   cmpPaste(x, y) 29.121 29.9050 32.22717 30.2515 31.0400 171.343   100
#  cld
#   a 
#   ab
#   ab
#    b
``

File: src/closest.c (hunk 4 above)
Comment (sgibb) [reply src/closest.c:81]: This may be over-engineered but this is needed to support *long vectors*:
(from `?"long vectors"`, highlighting by myself)
> Prior to R 3.0.0, all vectors in R were restricted to at most **2^31 - 1
> elements** and could be **indexed by integer** vectors.

> Currently all atomic (raw, logical, integer, numeric, complex,
> character) vectors, lists and expressions can be much longer on
> 64-bit platforms: such vectors are referred to as ‘long vectors’
> and have a slightly different internal structure. In theory they
> can contain up to **2^52 elements**, but address space limits of
> current CPUs and OSes will be much smaller.  Such objects will
> have a length that is expressed as a double, and can be **indexed by
> double** vectors.

So the question is: Is supporting *long vectors* (spectra with more than `2^31 - 1` elements) needed/useful at all.

(Just as anecdote: in `MALDIquant` I had `integer` for intensity values but were hit by integer overflows with my second data set (that contained way more peaks) for most operations (like TIC, noise estimation, ...). So I had to switch to `double` for intensity values. But currently I can't really imagine to hit the `2^31` limit with a single spectrum, but maybe with a library?)

**Comment by sgibb** [comment]
> Also, how are we going to deal with `NA` values? What would happen if we had an `NA` in either `x` or `table`?

`NA` in `table` isn't supported by `findInterval`. So it was not supported by `closest` either (we silently removed any `NA` before running the algorithm). For `NA` in `x` `NA` was returned. This will slightly increase the runtime of the algorithm but I could add an `ISNAN(px)` in the outer `for` loop and return `nomatch` for it. But is that a real use case? We should not have any `NA` in spectra mz values. And if there are any `NA` we could simply filter them with `x[!is.na(x)]` before calling `closest` (and avoid theses checks in the `for` loop (our main use case would be spectra matching which should be fast (and never contain any `NA` so the check would be a useless waste of runtime)).

File: R/matching.R (hunk 1 above)
Comment (jorainer) [reply R/matching.R:120]: OK, I see - but then, are you not missing a closing bracket? i.e. `!identical(FALSE, is.unsorted(x)) ||` instead of `!identical(FALSE, is.unsorted(x) ||`?

File: R/matching.R (hunk 2 above)
Comment (jorainer) [reply R/matching.R:133]: I don't think that it will brake anything at present - the *default* use case is to have a `tolerance` of length 1 and a `ppm` of length 1 - supporting `tolerance` of length > 1 is a *bonus*. So I think it is safe to change, but of course we can also leave it as it is. I have no strong opinion on that.

File: R/matching.R (hunk 3 above)
Comment (jorainer) [reply R/matching.R:139]: well, then let's keep the `if/else`.

File: src/closest.c (hunk 4 above)
Comment (jorainer) [reply src/closest.c:81]: hm, that's a valid question. I can not quite imagine having a spectrum with that many data points - at least not for LC-MS applications. Also, libraries (those that we use for metabolomics) usually have *clean* spectra with only few peaks in it - if you have 100 it's already a lot.

there's also the higher memory demand of double over integer - but that should not be a real problem here. So, I'm fine with either solution but somehow would prefer integer - but you can also keep `double`, but then you should update the help page saying that `closest` returns a `numeric` and not an `integer`.

**Comment by jorainer** [comment]
OK, I agree on the `NA`. Let's just assume that neither `x` nor `table` have `NA`s in it. But then this should also be clearly stated in the documentation.

File: R/matching.R (hunk 1 above)
Comment (sgibb) [reply R/matching.R:120]: Thanks, indeed.

File: src/closest.c (hunk 4 above)
Comment (sgibb) [reply src/closest.c:81]: Will change to `integer` and drop long vector support.

**Comment by jorainer** [comment]
Let me know when I should re-review.

File: R/matching.R (hunk 2 above)
Comment (sgibb) [reply R/matching.R]: I changed `tolerance` in relation to `x` now (`length(x) == length(tolerance)`).

**Comment by sgibb** [comment]
Please note that this PR doesn't fix #65 (it has the same behaviour as the former R implementation).

File: R/matching.R [hunk 5]
```diff
@@ -103,6 +106,10 @@ closest <- function(x, table, tolerance = Inf, ppm = 0,
                     duplicates = c("keep", "closest", "remove"),
                     nomatch = NA_integer_, .check = TRUE) {
 
+    ntolerance <- length(tolerance)
+    if (ntolerance != 1L && ntolerance != length(x))
```
jorainer, line 110: I'm wondering - should we put all these tests within a `if (.check)`? I guess there will only be a performance improvement if length of `tolerance` and or `ppm` is > 0.

jorainer (approved):
Thanks @sgibb - I just have on question. Apart from that I'm OK with merging.

R/matching.R
sgibb: I am not sure about this. I guess it won't decrease the runtime significantly.

PR #68, opened 2020-09-11
Add ... to join and similarity calculation functions

jorainer:
This is to fix issue https://github.com/rformassspectrometry/Spectra/issues/131


lgatto:
@jorainer - can you also push to Bioconductor, or do you want me to do it?

jorainer:
Can you please push to BioC?

lgatto:
Done - could you confirm that you have write access.

jorainer:
I think I don't need write access to the `MsCoreUtils` BioC repo (although I think I have) - it's enough if you and Sebastian can push there.

Issue #66, opened 2020-08-21
Problem with outer join if tolerance is large


jorainer:
I stumbled across the following problem:
performing an outer join on the two vectors below with a tolerance of 3 I would expect to have one match more than what we get:

```r
x <- c(1, 1.5, 2, 2.1, 5, 6, 7)
y <- c(4.6, 4.7, 4.8, 4.9, 5, 6, 7, 8)

MsCoreUtils:::.joinOuter(x, y, 3, 0)
$x
 [1]  1  2  3  4 NA NA NA NA  5  6  7 NA

$y
 [1] NA NA NA NA  1  2  3  4  5  6  7  8

```

In fact, I would expect to see a match also between `x[4]` and `y[1]` (i.e. `2.1` and `4.6`).
While this is a special case and in real-life spectra comparisons we might not have this case I would still want to fix that. I am currently checking if I could get it fixed in the C implementation of the outer join.

PR #69, opened 2020-09-23
refactor: linear search based closest C implementation

### LINKED ISSUE #65: Possible problem in left join matching if distances between two values are identical
The left join matching should find for each value in `x` the index of the value in `y` with the smallest difference (also given that the difference is smaller `tolerance`).

The current implementation of the left join (`MsCoreUtils:::.joinLeft`) seems to perform a non-intuitive match if the distance of a value in `x` to two values in `y` is the same.

Example:

```r
x <- as.numeric(c(1, 3, 5, 6, 8))
y <- as.numeric(c(3, 4, 5, 7))
```

with a `tolerance = 1` we have the following matches:
- `x[1]` (1): `NA`
- `x[2]` (3): matches both `y[1]` (3) and `y[2]` (4).
- `x[3]` (5): matches both `y[2]` (4) and `y[3]` (5).
- `x[4]` (6): matches both `y[3]` (5) and `y[4]` (7).
- `x[5]` (8): matches `y[4]` (7).

Now, given that we aim to find for each element in `x` the **closest** element in `y` we expect to get:

- `x[1]` - `NA`
- `x[2]` - `y[1]`
- `x[3]` - `y[3]`
- `x[4]` - `y[4]`
- `x[5]` - `NA`

The result of `MsCoreUtils:::.joinLeft(x, y, 1, 0)` is however:

```r
MsCoreUtils:::.joinLeft(x, y, 1, 0)
$x
[1] 1 2 3 4 5

$y
[1] NA  1  3 NA  4

```

i.e. it does not match `x[4]` (6) to `y[4]` (7) which would be the first match, but matches `x[5]` (8) to `y[4]` (7). While this is not wrong, I would expect from a left join to match `x[4]` to `y[4]`, i.e. that it uses the **first** match for each `x` in case there are ties.

sgibb:
This PR is a rewrite of the `closest` C implementation. Instead of the original binary it uses a linear search which is faster in our use cases (relatively small number of possible matches (< 1e7) and both arrays are sorted). Additionally it includes more validation checks in `if (.check)`. The implementation of `C_closest_dup_closest` is heavily inspired by @jorainer implementation of the left join in #63 .

This PR closes #65 and #66 for `closest` (however, we will replace the R implementation of the join* functions anyway).

BTW: the build on travis is failing with
```
--- re-building ‘MsCoreUtils.Rmd’ using rmarkdown
Quitting from lines 17-18 (MsCoreUtils.Rmd) 
Error: processing vignette 'MsCoreUtils.Rmd' failed with diagnostics:
argument is of length zero
--- failed re-building ‘MsCoreUtils.Rmd’
```
https://github.com/rformassspectrometry/MsCoreUtils/blob/eeec49f60bfc36e5dc7498c9efc34d98300ac2d2/vignettes/MsCoreUtils.Rmd#L17

I can't reproduce this locally. Any ideas?

### DIFF
--- a/src/closest.c
+++ b/src/closest.c
@@ -68,13 +68,14 @@ SEXP C_closest_dup_keep(SEXP x, SEXP table, SEXP tolerance, SEXP nomatch) {
  * \return index the closest element
  *
  * \note x and table have to be sorted increasingly and not containing any NA.
+ * \author Sebastian Gibb and Johannes Rainer
  */
 SEXP C_closest_dup_closest(SEXP x, SEXP table, SEXP tolerance, SEXP nomatch) {
     double *px = REAL(x);
     const unsigned int nx = LENGTH(x);
 
     double *ptable = REAL(table);
-    const unsigned int ntable1 = LENGTH(table) - 1;
+    const unsigned int ntable = LENGTH(table);
 
     double *ptolerance = REAL(tolerance);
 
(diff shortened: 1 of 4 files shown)


src/closest.c
```diff
@@ -70,91 +20,41 @@ int leftmost(double x, double* ptable, int low, int n) {
...
+
+        /* fabs should be just needed for the first element */
+        prevdiff = fabs(px[i] - ptable[j - 1]);
+        nextdiff = fabs(ptable[j] - px[i]);
+
+        if (prevdiff <= ptolerance[i] || nextdiff <= ptolerance[i]) {
+            if (prevdiff <= nextdiff)
+                pout[i] = j;
+            else
+                pout[i] = ++j;
```
jorainer, line 51: just a curiosity - is `j++` different than `++j`?

jorainer (approved):
Excellent PR @sgibb !

Just a curiosity: what is (if any) the performance gain compare to the original R implementations?

jorainer:
One thing maybe worthwhile looking into @sgibb: 

```
Warning: Found the following significant warnings:
  closest.c:225:13: warning: using floating point absolute value function 'fabs' when argument is of integer type [-Wabsolute-value]
```

src/closest.c
sgibb: `j++` increments `j` but returns `j` (reading as returning `j` and incrementing it afterwards) vs `++j` increments `j` and returns `j + 1` (reading as incrementing first and returning the result). So `pout[i] = ++j` is `j = j + 1; pout[i] = j`. `pout[i] = j++` would be `pout[i] = j; j = j + 1;`.

sgibb:
Thanks for your review @jorainer ! Here the missing benchmark against the latest pure R implementation:

``` r
library("MsCoreUtils")
#> 
#> Attaching package: 'MsCoreUtils'
#> 
#>     smooth
library("microbenchmark")
```
<details>
<summary>`closerR` implementation</summary>

```r
# from https://github.com/rformassspectrometry/MsCoreUtils/commit/3d371d9712b3e3083c6d3c0d8043c5c9f6644867
closestR <- function(x, table, tolerance = Inf, ppm = 0,
                    duplicates = c("keep", "closest", "remove"),
                    nomatch = NA_integer_) {

    if (!is.numeric(tolerance) || any(tolerance < 0))
        stop("'tolerance' has to be a 'numeric' larger or equal zero.")

    if(!is.numeric(ppm) || any(ppm < 0))
        stop("'ppm' has to be a 'numeric' larger or equal zero.")

    if (!is.numeric(nomatch) || length(nomatch) != 1L)
        stop("'nomatch' has to be a 'numeric' of length one.")

    table <- table[!is.na(table)]
    ntable <- length(table)
    if (!ntable)
        return(rep_len(nomatch, length(x)))

    ntolerance <- length(tolerance)

    if (ntolerance != 1L && ntolerance != ntable)
        stop("'tolerance' hat to be of length 1 or equal to 'length(table)'")

    tolerance <- tolerance + ppm(table, ppm) + sqrt(.Machine$double.eps)
    duplicates <- match.arg(duplicates)

    lIdx <- findInterval(x, table, rightmost.closed = FALSE, all.inside = TRUE)
    rIdx <- lIdx + 1L

    lIdx[lIdx == 0L] <- 1L
    rIdx[rIdx > ntable] <- ntable

    lDiff <- abs(table[lIdx] - x)
    rDiff <- abs(table[rIdx] - x)

    lDiff[lDiff > tolerance[lIdx]] <- Inf
    rDiff[rDiff > tolerance[rIdx]] <- Inf

    d <- which(lDiff > rDiff)

    lIdx[d] <- rIdx[d]

    ## no match at all
    lIdx[is.infinite(lDiff) & is.infinite(rDiff)] <- NA_integer_

    ## duplicated matches
    if (duplicates == "remove") {
        lIdx[duplicated(lIdx) | duplicated(lIdx, fromLast = TRUE)] <-
            NA_integer_
        lIdx[is.finite(lDiff) & is.finite(rDiff)] <- NA_integer_
    } else if (duplicates == "closest") {
        ## lIdx could be updated
        o <- order(abs(table[lIdx] - x))
        m <- lIdx[o]
        m[duplicated(m)] <- NA_integer_
        lIdx[o] <- m
    }

    if (!identical(nomatch, NA_integer_))
        lIdx[is.na(lIdx)] <- nomatch

    as.integer(lIdx)
}
```
</details>

```r
set.seed(42)
x <- sort(sample(1000, 700))
y <- sort(sample(1000, 700))
microbenchmark(
    cR=closestR(x, y, 0, 0, duplicates = "keep"),
    cT=closest(x, y, 0, 0, .check = TRUE, duplicates = "keep"),
    cF=closest(x, y, 0, 0, .check = FALSE, duplicates = "keep"),
    times = 1000
)
#> Unit: microseconds
#>  expr    min      lq      mean  median       uq       max neval cld
#>    cR 83.647 87.8935 168.49874 90.6410 136.6380 51880.921  1000   b
#>    cT 20.823 22.6280  28.43040 24.8855  33.9215    84.851  1000  a 
#>    cF 16.764 18.0115  23.08113 20.0490  26.8800   273.129  1000  a

microbenchmark(
    cR=closestR(x, y, 0, 0, duplicates = "closest"),
    cT=closest(x, y, 0, 0, .check = TRUE, duplicates = "closest"),
    cF=closest(x, y, 0, 0, .check = FALSE, duplicates = "closest"),
    times = 1000
)
#> Unit: microseconds
#>  expr     min       lq      mean   median       uq      max neval cld
#>    cR 162.742 170.4750 191.87221 173.4175 180.4795 4340.419  1000   b
#>    cT  21.531  23.9805  33.74464  27.6835  33.2360 4140.257  1000  a 
#>    cF  17.669  19.5565  28.58397  22.4390  27.5085 4076.893  1000  a

microbenchmark(
    cR=closestR(x, y, 0, 0, duplicates = "remove"),
    cT=closest(x, y, 0, 0, .check = TRUE, duplicates = "remove"),
    cF=closest(x, y, 0, 0, .check = FALSE, duplicates = "remove"),
    times = 1000
)
#> Unit: microseconds
#>  expr     min       lq      mean   median       uq      max neval cld
#>    cR 131.058 136.2625 162.48434 138.6580 142.9415 3683.219  1000   b
#>    cT  21.402  23.3755  27.27926  25.5750  28.5840   74.969  1000  a 
#>    cF  17.370  18.7665  21.87861  20.3565  23.3975   57.940  1000  a
```

<sup>Created on 2020-09-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

jorainer:
Amazing! Thanks @sgibb !

PR #73, opened 2020-10-23
add comment related to normalized dot product

jmbadia:
docs: add warning related to conceptual differences on normalized dot product method. 
@jorainer, I guess that is all. Please also let me know any improvement in my workflow


jorainer (approved):
Thank you @jmbadia for the update/corrections in the documentation!

Issue #75, opened 2020-10-28
formatRt() function


lgatto:
The `MSnbase::formatRt()` function is quite handy to format retention times from seconds (numeric) to/from a `mm:ss` character. 

Ok to move it to `MsCoreUtils` @jorainer @sgibb ?

jorainer:
Sure, OK for me.

lgatto:
See PR #76

sgibb:
To be honest I don't like that the function called `format` but converts rt in a way that depends on the type of its input.

What about two functions named e.g. `rt2character` and `rt2numeric`.

lgatto:
`formatRt()` has been transferred from `MSnbase` and I feel inclined to keep it. I think it is just simple from a user perspective but I appreciate that from a developer's perspective,  `rt2character()` and `rt2numeric()` are more straight to the point. 


We could have `rt2character()` and `rt2numeric()` and let `formatRt()` use these, with an error if the input is anything but a numeric or a character (as per your comment in #76).

lgatto:
See PR #77 for the suggestion above.

PR #76, opened 2020-10-29
Add formatRt - see #75

lgatto:


R/formatRt.R
```diff
@@ -0,0 +1,36 @@
...
+##' @return A reformatted retention time.
+##' 
+##' @author Laurent Gatto
+##'
+##' @export
+##'
+##' @examples
+##' formatRt(1524)
+##' formatRt(1)
+##' formatRt("25:24")
```
jorainer, line 17: I was wondering, does this also support `rt` to be of length > 1?

jorainer (commented):
Just one comment for now - whether the function is vectorized.

R/formatRt.R
lgatto: It does:
```r
> formatRt(1:10)
 [1] "0:01" "0:02" "0:03" "0:04" "0:05" "0:06" "0:07" "0:08" "0:09" "0:10"
```
I'll mention it in the man page and in the example.

lgatto: Addressed in latest addition to this PR.

In addition, I also now require an input of length > 0.

jorainer (approved):
I'm OK with this. Thanks @lgatto

R/formatRt.R
```diff
@@ -0,0 +1,39 @@
...
+##' formatRt(1:10)
+##' formatRt("25:24")
+##' formatRt(c("25:24", "25:25", "25:26"))
+formatRt <- function(rt) {
+    stopifnot(length(rt) > 0)
+    ans <- NA
+    if (is.numeric(rt)) {
+        min <- floor(rt / 60)
+        sec <- round(rt - (min * 60))
+        sec <- sprintf("%02d", sec)
```
sgibb, line 27: Why not just:
```r
ans <- sprintf("%d : %02d", min, sec)
```
instead of `sprintf` + `paste`?

R/formatRt.R
```diff
@@ -0,0 +1,39 @@
...
+        ans <- paste(min, ":", sec, sep = "")        
+    } else if (is.character(rt)) {
+        ans <- strsplit(rt, ":")
+        ans <- sapply(ans, function(x) {
+            x <- as.numeric(x)
+            60 * x[1] + x[2]
+        })
+    } else {
+        warning("Input must be numeric of character.")
```
sgibb, line 36: Shouldn't this be an error?

R/formatRt.R
```diff
@@ -0,0 +1,39 @@
...
+    }
+    return(ans)
```
sgibb, line 38: `return` isn't needed here.

sgibb (commented):
Thanks for the PR @lgatto.

R/formatRt.R
lgatto: Yes, changing it, thanks.

R/formatRt.R
lgatto: To be discussed in #75

PR #78, opened 2020-12-18
feat: imputation with use-provided function

lgatto:


jorainer (approved):
Looks good! Thanks @lgatto

PR #80, opened 2021-02-26
feat: add GNPS spectra similarity score

jorainer:
- Add `join_gnps` to allow peak matching as done in GNPS.
- Add `gnps` to calculate GNPS scores (if peaks matched with `join_gnps`).

Moved the code from `Spectra` here. Unit tests are included to compare the results against *reference values* extracted from original GNPS.


R/gnps.R
```diff
@@ -0,0 +1,167 @@
+#' @title GNPS spectrum similarity scores
+#'
+#' @description
+#'
+#' The similarity score used in [GNPS](https://gnps.ucsd.edu/) considers, in
```
lgatto, line 5: This sentence could be rephrase, as it is isn't entirely clear

> #' The similarity score used in [GNPS](https://gnps.ucsd.edu/) considers, in
> #' addition to directly matching peaks between two spectra, also matches of
> #' peaks with differences in their m/z values that correspond to the difference
> #' of the spectras' precursor m/z. 

What I understand is that peak in spectra x and y are matched twice: (1) using a default approach such as ppm or tolerance and (2) using a fixed delta m/z that is defined by the difference in the two spectras precuror m/z values. If this is the correct interpretations, I would suggest something along the lines of

> The similarity score used in [GNPS](https://gnps.ucsd.edu/) matching peaks between two spectra directly using a user-defined ppm or tolerance as well as a fixed delta m/z that is defined by the difference in the two spectras precuror m/z values.

R/gnps.R
```diff
@@ -0,0 +1,167 @@
...
+#' addition to directly matching peaks between two spectra, also matches of
+#' peaks with differences in their m/z values that correspond to the difference
+#' of the spectras' precursor m/z. For peaks that match multiple peaks in the
+#' other spectrum only the matching peak pair with the higher value/similarity
+#' is considered in the final similarity score. Peak matching that considers
+#' also differences in precursor m/z can be performed with `join_gnps` and the
+#' similarity score on such *aligned* peak matrices can be performed with
+#' `gnps`. Note that only if the two functions are used together, GNPS
+#' similarity scores are calculated.
```
lgatto, line 14: Note that GNPS similarity scores are calculated only if the two functions are used together.

R/gnps.R
```diff
@@ -0,0 +1,167 @@
...
+#'
+#' - `join_gnps`: matches/maps peaks between spectra with the same approach
+#'   used in GNPS: peaks are considered matching if a) the
```
lgatto, line 17: as in GNPS

R/gnps.R
lgatto: The description below seems much better to me, and points to possible inaccuracies in my understanding. So please, amend as you see fit.

R/gnps.R
```diff
@@ -0,0 +1,167 @@
...
+#'   their m/z *adjusted* for the difference of the spectras' precursor is
+#'   smaller than defined by `tolerance` and `ppm`. Based on this definition,
+#'   peaks in `x` can match up to two peaks in `y` hence returned peak indices
+#'   might be duplicated. Note that if one of `xPrecursorMz` or `yPrecursorMz`
+#'   are `NA` or if both are the same, the results are the same as with
+#'   [join()]. The function returns a `list` of two `integer` vectors with the
+#'   indices of the peaks matching peaks in the other spectrum or `NA`
+#'   otherwise.
+#'
+#' - `gnps`: calculate the GNPS similarity score on peak matrices previously
```
lgatto, line 29: calculate**s**

lgatto, line 29: on matrices'

R/gnps.R
```diff
@@ -0,0 +1,167 @@
...
+#'   *aligned* (matched) with `join_gnps`. For multi-mapping peaks the pair with
+#'   the higher similarity are considered in the final score calculation.
+#'
+#' @details
+#'
+#' The implementation of `gnps` bases on the R code from the publication listed
```
lgatto, line 35: is based on Xing *et al.* (2020)

R/gnps.R
```diff
@@ -0,0 +1,167 @@
...
+#' ## with one of the two precursor m/z being NA, the result are the same as
+#' ## with join.
+#' join_gnps(x[, 1], y[, 1], pmz_x, yPrecursorMz = NA)
+#'
+#' ## Calculate GNPS similarity score:
+#' map <- join_gnps(x[, 1], y[, 1], pmz_x, pmz_y)
+#' gnps(x[map[[1]], ], y[map[[2]], ])
+gnps <- function(x, y, ...) {
+    if (nrow(x) != nrow(y))
+        stop("'x' and 'y' are expected to be aligned peak matrices (i.e. ",
```
lgatto, line 117: The test here seems to test the equality of rows only, and I would thus expect the error message to complain about that. Would there be another test regarding the aligned nature of the matrices?

lgatto (changes requested):

R/gnps.R
jorainer: Thanks. Yes that's indeed easier.

R/gnps.R
jorainer: OK, I will change the error message. I don't see a way how we could check that the peak matrices are indeed aligned.

jorainer:
Thanks for the fast review @lgatto ! I've updated the documentation accordingly.

lgatto (approved):

PR #88, opened 2021-11-13
New aggregate_by_matrix() function

lgatto:
New `aggregate_by_matrix()` that uses an adjacency matrix to aggregate quantitative features.


R/aggregate_by_matrix.R
```diff
@@ -0,0 +1,37 @@
+#' @rdname aggregate
+#'
+#' @export
+colMeansMat <- function(x, MAT)
+  colSumsMat(x, MAT)/colSums(MAT)
```
sgibb, line 5: Just styling
```suggestion
  colSumsMat(x, MAT) / colSums(MAT)
```

R/aggregate_by_matrix.R
```diff
@@ -0,0 +1,37 @@
...
+
+#' @rdname aggregate
+#'
+#' @export
+colSumsMat <- function(x, MAT)
+  t(MAT) %*% x
+
+
+##' @param MAT An adjacency matrix that defines what features with
```
sgibb, line 14: At least for me it is a little bit difficult to understand: "defines what features"? defines what elements are features?

R/aggregate_by_matrix.R
```diff
@@ -0,0 +1,37 @@
...
+##'     of dimensions `ncol(MAT)` and `ncol(x), with `dimnames` equal
+##'     to `colnames(x)` and `rownames(MAT)`.
+##'
+##' @rdname aggregate
+##'
+##' @export
+aggregate_by_matrix <- function(x, MAT, FUN, ...) {
+    if (!inherits(x, "matrix"))
+        stop("'x' must be a matrix.")
+    if (!inherits(MAT, "Matrix") & !inherits(MAT, "matrix"))
```
sgibb, line 27: You should use `&&` to skip this `if` statement if the first test is `TRUE` already, e.g. MAT is a `Matrix`.
```suggestion
    if (!inherits(MAT, "Matrix") && !inherits(MAT, "matrix"))
```
BTW: It would be good to set the most common class in front here. Do you think most user will use `Matrix` for `MAT`? Otherwise I would suggest to test for `matrix` first.

R/aggregate_by_matrix.R
```diff
@@ -0,0 +1,37 @@
...
+##'
+##' @return [aggregate_by_matrix()] returns a `matrix` (or `Matrix`)
+##'     of dimensions `ncol(MAT)` and `ncol(x), with `dimnames` equal
+##'     to `colnames(x)` and `rownames(MAT)`.
+##'
+##' @rdname aggregate
+##'
+##' @export
+aggregate_by_matrix <- function(x, MAT, FUN, ...) {
+    if (!inherits(x, "matrix"))
```
sgibb, line 25: Wouldn't be `is.matrix` sufficient here?

R/aggregate_by_matrix.R
sgibb, line 27: I'm often unsure when to use `inherits` or `is`. Isn't `Matrix` a *S4* class and should be tested with `is`?

R/aggregate_by_matrix.R
```diff
@@ -0,0 +1,37 @@
...
+        stop("'x' must be a matrix.")
+    if (!inherits(MAT, "Matrix") & !inherits(MAT, "matrix"))
+        stop("'MAT' must be a matrix.")
+    if (!identical(nrow(MAT), nrow(x)))
+        stop("nrow(MAT) must be identical to 'nrow(x).")
+    res <- do.call(FUN, list(x, MAT, ...))
+    if (!is.null(colnames(MAT)))
+        stopifnot(identical(colnames(MAT), rownames(res)))
```
sgibb, line 33: `stopifnot` is a slow function with multiple tests and an often unhelpful error message. That's why I would suggest the following:
```suggestion
    if (!is.null(colnames(MAT)) && !identical(colnames(MAT), rownames(res)))
        stop("The colum names of 'MAT' have to be identical to the row names of 'res'!")
```

R/aggregate_by_matrix.R
```diff
@@ -0,0 +1,37 @@
...
+    if (!is.null(colnames(x)))
+        stopifnot(identical(colnames(x), colnames(res)))
```
sgibb, line 35: Same as above:
```suggestion
    if (!is.null(colnames(x)) && !identical(colnames(x), colnames(res)))
        stop("The column names of 'x' have to be identical to the column names of 'res'!")
```

R/aggregate_by_vector.R
```diff
@@ -0,0 +1,173 @@
...
+##'
+##' @rdname aggregate
+##'
+##' @export
+aggregate_by_vector <- function(x, INDEX, FUN, ...) {
+    if (!inherits(x, "matrix"))
+        stop("'x' must be a matrix.")
+    if (!identical(length(INDEX), nrow(x)))
+        stop("The length of 'INDEX' has to be identical to 'nrow(x).")
+    INDEX <- factor(INDEX)
```
sgibb, line 162: I don't think this line is necessary, from `?tapply`:
> INDEX: a ‘list’ of one or more ‘factor’s, each of same length as
          ‘X’.  The elements are coerced to factors by ‘as.factor’.

R/aggregate_by_vector.R
```diff
@@ -0,0 +1,173 @@
...
+    res <- tapply(
+        seq_len(nrow(x)),
+        INDEX,
+        FUN = function(i) FUN(x[i, , drop = FALSE], ...),
+        simplify = FALSE
+    )
+    res <- do.call(rbind, res)
+    rownames(res) <- levels(INDEX)
```
sgibb, line 170: This line should also not be necessary, because `tapply` names the list elements by `INDEX` and `do.call(rbind, ...` will preserve them.

R/aggregate_by_vector.R
```diff
@@ -0,0 +1,173 @@
...
+    colnames(res) <- colnames(x)
```
sgibb, line 171: Same as above.

R/aggregate_by_vector.R
```diff
@@ -0,0 +1,173 @@
...
+        stop("'x' must be a matrix.")
+    if (!identical(length(INDEX), nrow(x)))
+        stop("The length of 'INDEX' has to be identical to 'nrow(x).")
+    INDEX <- factor(INDEX)
+    res <- tapply(
+        seq_len(nrow(x)),
+        INDEX,
+        FUN = function(i) FUN(x[i, , drop = FALSE], ...),
+        simplify = FALSE
+    )
```
sgibb, line 168: I would simplify this call by `lapply(split(...` (which is more or less what `tapply(..., simplify = FALSE)` is doing:
```suggestion
    FUN <- match.fun(FUN)
    res <- lapply(
        split(seq_len(nrow(x)), INDEX),
        FUN = function(i) FUN(x[i, , drop = FALSE], ...)
    )
```
It is less code and slightly improves the runtime:
```r
library("microbenchmark")
set.seed(1)

m <- matrix(sample(1000, 1000, repl = TRUE), nrow = 100)
i <- sample(100, 100, repl = TRUE)

aggregate_by_vector <- function(x, INDEX, FUN, ...) {
    FUN <- match.fun(FUN)
    res <- tapply(
        seq_len(nrow(x)),
        INDEX,
        FUN = function(i) FUN(x[i, , drop = FALSE]),
        simplify = FALSE
    )
    do.call(rbind, res)
}

aggregate_by_vector2 <- function(x, INDEX, FUN, ...) {
    fun <- match.fun(FUN)
    res <- lapply(
        split(seq_len(nrow(x)), INDEX),
        FUN = function(i) fun(x[i, , drop = FALSE])
    )
    do.call(rbind, res)
}

microbenchmark(
    tapply = aggregate_by_vector(m, i, colMeans),
    lapply = aggregate_by_vector2(m, i, colMeans),
    check = "identical"
)
# Unit: microseconds
#    expr     min      lq     mean  median       uq      max neval
#  tapply 516.905 553.142 621.4333 578.680 613.2660 2384.372   100
#  lapply 480.152 512.610 539.0358 531.765 560.4715  751.225   100
``

**Review by sgibb — CHANGES_REQUESTED** [review]
Dear Laurent, thanks for this nice PR. The aggregation by an adjacency matrix is a really nice feature. I have just a few minor suggestions to discuss.

File: R/aggregate_by_matrix.R [hunk 11]
```diff
@@ -0,0 +1,37 @@
...
+#'
+#' @export
+colMeansMat <- function(x, MAT)
+  colSumsMat(x, MAT)/colSums(MAT)
+
+#' @rdname aggregate
+#'
+#' @export
+colSumsMat <- function(x, MAT)
+  t(MAT) %*% x
```
sgibb, line 11: (maybe neglectable speed improvement); from `?crossprod`:
> This is formally equivalent to (but usually
     slightly faster than) the call ‘t(x) %*% y’ (‘crossprod’) or ‘x
     %*% t(y)’ (‘tcrossprod’).

```suggestion
  crossprod(MAT, x)
```

R/aggregate_by_matrix.R
lgatto: I think they would be equivalent here. I am just used to `is()` or `inherits()`. I don't have a strong opinion here, so happy to change.

R/aggregate_by_matrix.R
lgatto: I think that most uses will be `Matrix` (a sparse matrix, actually), as these adjacency matrices are likely to be created with [PSMatch::makeAdjacencyMatrix()](https://rformassspectrometry.github.io/PSMatch/reference/adjacencyMatrix.html). For large data, these matrices are much smaller and faster to generate. Hence the appearance of `Matrix` here.

lgatto: There is indeed a small difference:

> The analogue of ‘inherits’ for formal classes is is.  The two
     functions behave consistently with one exception: S4 classes can
     have conditional inheritance, with an explicit test.  In this
     case, is will test the condition, but ‘inherits’ ignores all
     conditional superclasses.

And `is()` is more appropriate here. Will change

R/aggregate_by_matrix.R
lgatto: Thanks - I suppose I just like writing `stopifnot()` :-)

R/aggregate_by_vector.R
lgatto: Ok, thank, will remove it.

R/aggregate_by_vector.R
lgatto: Can't argue with a benchmark!

R/aggregate_by_vector.R
lgatto: ok, to be removed, thanks

R/aggregate_by_vector.R
lgatto: ok

R/aggregate_by_matrix.R
lgatto: Indeed, part of that sentence got lost. Will amend.

lgatto:
Thank you very much @sgibb for your rigorous and high-quality review. I have implemented all your suggestions.

lgatto:
@sgibb - some changes re these two lines:
 
```r
INDEX <- as.factor(INDEX) 
```
and
```r
rownames(res) <- levels(INDEX)
```
are needed.

The second line is needed because if one of the levels has only one element, it's the rowname of that element that is used to set the name, rather than that level's name. Manually setting `as.factor(INDEX)` and then using the levels as names makes sure that the order is always correct. 

I'll revert these two lines back.

R/aggregate.R
```diff
@@ -1,222 +1,123 @@
...
-    ## Sort the sample coefficients in the same way as the samplenames
-    ## of expression matrix. Puts NA for the samples without any
-    ## expression value
-    res <- coef[paste0('sample', colnames(x))]
-    names(res) <- colnames(x)
-    res
-}
-
-##' @title Return the Median Polish (Robust Twoway Decomposition) of a matrix
+##' Adjacency matrices are an elegant way to explicitly encode for
```
jorainer, line 10: Just out of curiosity: what's the advantage of a `matrix`-like structure over a `list` of `integer` indices? I had the feeling that sparse matrices can be sometimes quite slow.

lgatto: For a real-life scenario of an adjacency matrix, see the [makeAdjacencyMatrix()](https://rformassspectrometry.github.io/PSMatch/reference/adjacencyMatrix.html) manual page. I compared the implementation of that function on an mzID file using a regular and a sparse matrix, and the latter was smaller (as expected) but the code was substantially faster (at least with my simple implementations). I think the difference can be explained by the way the matrices are created:

```r
> m <- matrix(0, nrow = 3, ncol = 2)
> m[1, 1] <- m[2, 1]  <- m[3, 2]  <- 1 ## actually a for loop 
> 
> m
     [,1] [,2]
[1,]    1    0
[2,]    1    0
[3,]    0    1
> library(Matrix)
> sparseMatrix(i = c(1, 2, 3), j = c(1, 1, 2), x = 1)
3 x 2 sparse Matrix of class "dgCMatrix"
        
[1,] 1 .
[2,] 1 .
[3,] . 1
```
I also tested a couple of relevant cross products with these adjacency matrices, and sparse matrices were efficient.

jorainer: thanks for the clarification

PR #90, opened 2021-12-03
fix: set colnames after aggregation

lgatto:
This PR makes sure that the colnames are always present after aggregation:

Current behaviour is 

```r
> m <- matrix(1:8, ncol = 2, dimnames = list(letters[1:4], paste0("S", 1:2)))
> m
  S1 S2
a  1  5
b  2  6
c  3  7
d  4  8
> ## below reproduces aggregate_by_vector()
> res <- lapply(split(seq_len(nrow(m)), rep(1:2, each = 2)), FUN = function(i) colMedians(m[i, , drop = FALSE]))
> nms <- names(res)
> res <- do.call(rbind, res)
> rownames(res) <- nms
> res
  [,1] [,2]
1  1.5  5.5
2  3.5  7.5
````
New/desired behaviour:
```r
> colnames(res) <- colnames(m)
> res
   S1  S2
1 1.5 5.5
2 3.5 7.5
```

Note that this isn't an issue for every aggregation function:
```r
> res <- lapply(split(seq_len(nrow(m)), rep(1:2, each = 2)), FUN = function(i) colMeans(m[i, , drop = FALSE]))
> nms <- names(res)
> res <- do.call(rbind, res)
> rownames(res) <- nms
> res
   S1  S2
1 1.5 5.5
2 3.5 7.5
```

As for the `colMedians()`, there is a `useNames = TRUE` argument to fix it, but this should be handled consistently as part of the `MsCoreUtils::aggregate_by_vector()` and `MsCoreUtils::aggregate_by_matrix()` functions. 

This was caught by a unit test that worked before, so I suspect that way `colMedians()` deals with the names has changed.


jorainer (approved):
Thanks for the PR @lgatto . Fine with me.

PR #91, opened 2021-12-09
Use `Matrix::colSums()` by default

lgatto:
In this PR, I suggest to use `Matrix::colSums()` by default, whether the matrix is a `matrix` or a `Matrix`. 

An alternative would of course be to check the class of the variable and use `base::colSums()` or `Matrix::colSums()` accordingly. Personally, I think that this adds annoying (and unnecessary) code when (1) most actual cases will be sparse matrices (so not `matrix`) and (2) I would be tempted to assume that instances of `[dgC]Matrix` will indeed be used and simplify the code based on that assumption, and (3) if we encounter a `matrix`, we convert it early on to a `sparseMatrix`. This doesn't add any actual dependency as `Matrix` is a recommended package anyway. 

@sgibb @jorainer, what do you think?


lgatto:
@jorainer and @sgibb - I need to merge this PR as `QFeatures` depends on this minor change but have opened #92 to discuss further and adapt based on your thoughts.

Issue #94, opened 2022-02-22
closest not working with negative m/z values


jorainer:
With positive values all works fine:

```r
library(MsCoreUtils)
a <- c(35, 36, 37)
closest(36.01, a, tolerance = 0.1)
[1] 2

closest(36.00036, a, tolerance = 0, ppm = 15)
[1] 2
```

But with negative ones we fail to detect the match with `ppm`:

```r
closest(-36.01, sort(-a), tolerance = 0.1)
[1] 2

closest(-36.00036, sort(-a), tolerance = 0, ppm = 15)
[1] NA
```

This might in fact just be a missing `abs` somewhere within the code.

lgatto:
Rather than a missing `abs()`, shouldn't this rather throw an error due to negative M/Z values?

jorainer:
I would like to support also negative m/z values - this comes from the idea of the *neutral loss* spectra where the fragment peaks' m/z values are substracted from the precursor m/z (i.e. the m/z reported in the resulting spectra are precursor m/z - fragment m/z). It can happen that the fragment m/z is larger than the precursor m/z and in these cases the resulting m/z would be negative.

Apart, I think the `closest` should not only work for m/z values but any type of values, also negative ones.

PR #95, opened 2022-02-22
fix: fix issue #94

jorainer:
- Support matching of negative values in `closest`, `group` and `join` with ppm
  specified.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: MsCoreUtils
 Title: Core Utils for Mass Spectrometry Data
-Version: 1.7.3
+Version: 1.7.4
 Description: MsCoreUtils defines low-level functions for mass
     spectrometry data and is independent of any high-level data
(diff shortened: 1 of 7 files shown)


lgatto (approved):

sgibb:
While this PR is correct I am wondering if we fix it on the wrong place. Shouldn't `ppm` always return a positive value? Is `ppm(-1000, 15) == -0.015` correct/expected? What's about `ppm(1000, -15)`? Is that even useful?

IMHO we may should think about replacing the current `ppm` with `ppm <- function(x, ppm)abs(x * ppm * 1e-06)`.

jorainer:
Thanks @sgibb , yes, I think you're right. The `ppm` (AFAIK) should always be positive. I will change that and clearly state it also in the documentation. Regarding `ppm(1000, -19)`, that does not make sense to me - we check that also in the `closest` et al functions.

PR #99, opened 2022-09-28
Provide compatibility with HDF5Matrix

cvanderaa:
Hello, 

This PR provides support for matrices stored in HDF5. This is in response to an ongoing issue in `QFeatures`: https://github.com/rformassspectrometry/QFeatures/issues/171

There are several functionality in `MsCoreUtils` that do not allow to provide matrices in an other class than `matrix`. The functionality I currently identified to be adapted (for `QFeatures`) are aggregation, imputation and normalization. 

### Discussion

**please do not merge!** In the current state of the PR, I adapted the code only for aggregation as an example. I would like to discuss several points: 

1. Do you would agree on having HDF5 support for `MsCoreUtils`?
2. If yes, do you agree on the implementation so far?
3. I face one important consideration. Given an `HDF5Matrix` object as input, should we expect an `HDF5Matrix` object as output? For the moment, I convert to `matrix` for convenience so that `aggregate_by_{matrix|vector}()` systematically return a `matrix`. 
4. Later on, I can work on imputation and normalization. Can you think of other steps that should be adapted?

### bonus

I found a little bug in `aggregate_by_matrix()` when missing data is present. I solved the issue and also include an `na.rm` argument as suggested in the example. I can make a separate PR if needed.

### DIFF
--- a/R/aggregate.R
+++ b/R/aggregate.R
@@ -8,9 +8,7 @@
 ##' method is defined by function `FUN`.
 ##'
 ##' Adjacency matrices are an elegant way to explicitly encode for
-##' shared peptides (see example below) during aggregation. Note
-##' however that missing values can't generally be ignore using `na.rm
-##' = TRUE`, as with vector-based aggregation (see examples below).
+##' shared peptides (see example below) during aggregation.
 ##'
 ##' @section Vector-based aggregation functions:
 ##'
@@ -43,6 +41,14 @@
 ##'    peptide intensities. Shared peptides are re-used multiple
 ##'    times.
 ##'
+##' @section Handling missing values:
+##' 
+##' By default, missing values in the quantitative data will propagate
+##' to the aggregated data. You can provide `na.rm = TRUE` to most 
+##' functions listed above to ignore missing values, except for 
+##' `robustSummary()` where you should supply `na.action = na.omit` 
+##' (see `?MASS::rlm`). 
+##' 
 ##' @family Quantitative feature aggregation
 ##'
 ##' @name aggregate
--- a/R/aggregate.R
+++ b/R/aggregate.R
@@ -100,22 +100,17 @@
 ##'                             c("S1", "S2")))
 ##' x
 ##'
-##' ## simply use na.rm = TRUE to ignore missng values
+##' ## simply use na.rm = TRUE to ignore missing values
 ##' ## during the aggregation
 ##'
 ##' (k <- LETTERS[c(1, 1, 2)])
 ##' aggregate_by_vector(x, k, colSums)
 ##' aggregate_by_vector(x, k, colSums, na.rm = TRUE)
 ##'
-##'
-##' ## NAs are propagated during the matrix
-##' ## multiplication in the aggregation functon
 ##' (adj <- matrix(c(1, 1, 0, 0, 0, 1), ncol = 2,
 ##'                dimnames = list(paste0("Pep", 1:3),
-##'                                c("A", "B"))))
-##'
-##' aggregate_by_matrix(x, adj, colSumsMat)
-##' ## not implemented
-##' try(aggregate_by_matrix(x, adj, colSumsMat, na.rm = TRUE))
-##' colSumsMat
+##'                            c("A", "B"))))
+##' aggregate_by_matrix(x, adj, colSumsMat, na.rm = FALSE)
+##' aggregate_by_matrix(x, adj, colSumsMat, na.rm = TRUE)
+##' 
 NULL
--- a/tests/testthat/test_aggregate.R
+++ b/tests/testthat/test_aggregate.R
@@ -96,12 +96,16 @@ test_that("aggregation: aggregate_by_vector", {
     expect_equal(x2_robust_expected[same_row_names, ],
                  aggregate_by_vector(x, k_fact2, robustSummary)[same_row_names, ])
     ## Test HDF5Matrix compatibility
+    hdf5out <- aggregate_by_vector(xhdf5, k_char, robustSummary)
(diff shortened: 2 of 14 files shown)


jorainer:
Thanks for the PR Christophe!

Some of my thoughts (haven't had a close look at the code yet):

1) I would like to keep the dependencies of the `MsCoreUtils` as low as possible. We have many packages importing from `MsCoreUtils`, so I would prefer to have it as lightweight as possible.

2) Looking at the code I'm a bit puzzled - you're actually not importing anything from `HDF5Matrix` - all the calls I see are `Matrix::`. I suppose that `HDF5Matrix` extends `Matrix`? Do you actually need to put  `HDF5Matrix` into `Suggests`? Should also work without that - if you only use `HDF5Matrix` in the documentation or unit tests you can also put it into `Enhances`. That could actually resolve the point 1) above. If the user provides an `HDF5Matrix` as input he/she will have loaded the package anyway, so the namespace (if it's using S4 classes) should be available for `MsCoreUtils` at that stage anyway.

3) I would suggest the functions that work on matrices and return matrices to return the same type of matrix: `matrix` as input, `matrix` as output, `Matrix` as input `Matrix` as output.

cvanderaa:
I adapted the points discussed above. I only face an issue with the `Enhances` field. It seems `HDF5Array` is no longer accessible for the unit tests.

I would love to get a code review.

tests/testthat/test_aggregate.R
```diff
@@ -1,3 +1,5 @@
+library(HDF5Array)
```
jorainer, line 1: instead of adding this here I would suggest to require the namespace for the package only in the unit tests were you use this package. See for example: [here](https://github.com/rformassspectrometry/MetaboAnnotation/blob/master/tests/testthat/test_CompDbSource.R#L43)

Also, I would then maybe manually install the `HDF5Array` in the github action (using `BiocManager::install("HDF5Array")`.

jorainer:
I'll try to have a look at the code over the weekend. In the meantime you could maybe try to fix the GH error on Windows with my suggestion in the comment above.

R/aggregate.R
```diff
@@ -100,22 +100,17 @@
...
 ##' (adj <- matrix(c(1, 1, 0, 0, 0, 1), ncol = 2,
 ##'                dimnames = list(paste0("Pep", 1:3),
-##'                                c("A", "B"))))
-##'
-##' aggregate_by_matrix(x, adj, colSumsMat)
-##' ## not implemented
-##' try(aggregate_by_matrix(x, adj, colSumsMat, na.rm = TRUE))
-##' colSumsMat
+##'                            c("A", "B"))))
+##' aggregate_by_matrix(x, adj, colSumsMat, na.rm = FALSE)
```
jorainer, line 113: I find the documentation a bit confusing as it is stated that `na.rm` is not used. Could you maybe expand/fix the documentation and clearly state when `na.rm` can be used and where not?

tests/testthat/test_aggregate.R
```diff
@@ -130,7 +129,48 @@ test_that("aggregation: aggregate_by_vector", {
                  aggregate_by_vector(x, k_fact, colSums)[same_row_names, ])
     expect_equal(x2_sum_expected[same_row_names, ],
                  aggregate_by_vector(x, k_fact2, colSums)[same_row_names, ])
-
+
```
jorainer, line 132: would it maybe make sense to test also on `Matrix`?

jorainer (changes requested):
Thanks Christophe for the PR!

I only have some small comments/requests mostly on the documentation and unit tests.

tests/testthat/test_aggregate.R
cvanderaa: Sure, `Matrix` would make sense as well, although we are not very interested in this feature for `QFeatures` application (cf: https://github.com/rformassspectrometry/QFeatures/issues/157). Do you want me to provide compatibility for `Matrix` objects as well?

jorainer: It was more just to have also a unit test that runs on systems that don't have HDF5Array available - sort as a proof of working functionality. Makes however only sense if the HDF5Array extends `Matrix`. If not, just ignore this comment.

cvanderaa: Ok, I misunderstood your question. I don't think this would work since `inherits(hdf5MatrixObject, "Matrix")` returns false so it does not extend `Matrix`. But I get your point and I can't think of a solution that would test the chunks in the absence of `HDF5Array`. The only improvement I could see is to use `testthat::skip_if()` to at least throw a message that the chunk was not tested. What do you think?

jorainer: Ah, no, then ignore my comment. All good.

jorainer (approved):
Looks good to me! Thanks!

cvanderaa:
The PR now enables HDF5 compatibility for aggregation, imputation and normalization. @lgatto would you mind a final review before merging? 

One important point that I currently leave aside is that the functions are compatible for aggregation, imputation and normalization, but are far from optimized. In fact, I convert the `HDF5Matrix` object to `matrix` at the start of the function and save a new `HDF5` dataset with the processed matrix at the end of the function. A cleverer implementation would define new functions specifically for `HDF5Matrix` objects and handle block chunking using `DelayedArray::blockApply()`. This will ask quite some time to implement because the chunking will depend on the normalization or imputation method used.  If needed I can do this in a new PR, but that will for sure not be ready before Bioc3.17 release.

lgatto (approved):

lgatto:
Fine by me, thank you @cvanderaa !

PR #104, opened 2022-11-23
DescendPeaks in C++

Pascallio:
This is the C++ implementation of the descendPeaks function of MSnbase. It takes a slightly different approach to the original by first subsetting the maximum region to consider. Next for each left/right descend, it checks if the next signal is A) above the signal percentage threshold and B) not higher than the previous signal (rising). If the signal passes both tests, the left/right index is updated. After both descends, these indexes are used to define the region of which the weighted mean of the mass is taken.

It is recommended to sanatize all inputs first in R before calling this function to ensure the right values are used, e.g. C-indexes, not R-indexes.

### DIFF
--- a/src/descendPeak.cpp
+++ b/src/descendPeak.cpp
@@ -16,10 +16,8 @@ double weightedMean(NumericVector mzs, NumericVector ints) {
 }
 
 double descend(int centroid, NumericVector intensities,
-               NumericVector masses, double signalRatio) {
+               NumericVector masses, double signalRatio, int maxK = 30) {
 
-    // Maximum deviation from the centroid index
-    int maxK = 30;
 
     // Define a min and max index for the region where the region is not negative
     // and does not exceed the size of the intensities vector, since we only
@@ -124,7 +122,7 @@ double descend(int centroid, NumericVector intensities,
 //' @export
 // [[Rcpp::export]]
 NumericVector descendPeaksC(NumericVector centroids, NumericVector intensities,
-                           NumericVector masses, double signalPercentage) {
+                           NumericVector masses, double signalPercentage, int maxK = 30) {
 
     // Prepare variables
     int n = centroids.length();
@@ -135,7 +133,7 @@ NumericVector descendPeaksC(NumericVector centroids, NumericVector intensities,
     for (int i = 0; i < n; i++) {
         // Calculate the weighted mass
         weightedMZs[i] = descend(centroids[i], intensities,
-                                 masses, signalRatio);
+                                 masses, signalRatio, maxK);
     }
     return weightedMZs;
 }
--- a/src/descendPeak.cpp
+++ b/src/descendPeak.cpp
@@ -1,0 +1,143 @@
+// Pascal Maas - p.maas@lacdr.leidenuniv.nl
+
+#include <Rcpp.h>
+using namespace Rcpp;
+using namespace std;
+
+double weightedMean(NumericVector mzs, NumericVector ints) {
+    // Get the total intensity
+    double intensity = sum(ints);
+
+    // Calculate the weighted mass
+    int n = ints.length();
+    double total = 0;
+    for (int i = 0; i < n; i++) {
+        total += mzs[i] * ints[i] / intensity;
+    }
+    return total;
+}
+
+double descend(int centroid, NumericVector intensities,
+               NumericVector masses, double signalRatio) {
+
+    // Maximum deviation from the centroid index
+    int maxK = 30;
+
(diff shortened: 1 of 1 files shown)


lgatto:
Just a quick comment - it would also be important to have unit tests to check that the results give expected results on small data or/and same as an R implementation on a more complex data.

src/descendPeak.cpp
```diff
@@ -0,0 +1,143 @@
...
+    double intensity = sum(ints);
+
+    // Calculate the weighted mass
+    int n = ints.length();
+    double total = 0;
+    for (int i = 0; i < n; i++) {
+        total += mzs[i] * ints[i] / intensity;
+    }
+    return total;
+}
```
sgibb, line 18: Without testing I would assume that you could reduce the runtime of this function by avoid looping over the vector twice (sum has to go through all indices as well):
```suggestion
double weightedMean(NumericVector mzs, NumericVector ints) {
    int n = ints.length();
    double total = 0;
    double intensity = 0;
    for (int i = 0; i < n; i++) {
        total += mzs[i] * ints[i];
        intensity += ints[i];
    }
    return total / intensity;
}
```

sgibb:
@Pascallio Thanks for your contribution. It is very welcome! Maybe a stupid question/suggestion. We try to keep the dependencies as low as possible (Suggests contains already a large list of dependencies). Currently we successfully avoid `Rcpp` as direct dependency (some packages in Suggests of course already depend on `Rcpp`). It would be great if you ensure that `Rcpp` would end in `Suggests` and not in `Depends` or `Imports` or (implement this whole stuff in pure C).

src/descendPeak.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+double descend(int centroid, NumericVector intensities,
+               NumericVector masses, double signalRatio) {
+
+    // Maximum deviation from the centroid index
+    int maxK = 30;
```
sgibb, line 24: Depending on the solution of the spectra an window of +/- 30 may be to large/small. I don't like to hardcode window sizes. I would suggest to make it an argument as well.

src/descendPeak.cpp
```diff
@@ -0,0 +1,143 @@
...
+    for (int i = centroid - 1; i >= 0; i--) {
+        // Get current and previous values
+        double previous = ints[i + 1];
+        double current = ints[i];
+
+        // Good signal should be above the centroid / signal ratio
+        bool goodSignal =  current / centroidValue > signalRatio;
+
+        // Compare current to previous value to check for a rising signal
+        bool risingSignal = current >= previous;
```
sgibb, line 65: Would this be a good test? Already a small plateau (2 indices of the same intensity value) would break here, even if it goes down after the overnext index.

src/descendPeak.cpp
Pascallio: I agree, this improves the weighted mean function.

src/descendPeak.cpp
Pascallio: Agreed, I will push an update that adds the `maxK` parameter to the function arguments.

src/descendPeak.cpp
Pascallio: I do believe that is the default behaviour in the original MSnbase version here: https://github.com/lgatto/MSnbase/blob/1338323345d2a6cf856f1819dfb365300ab6d9d5/R/functions-Spectrum.R#L781-L788

I'm not sure about the current version in MsCoreUtils, but I could implement the `stopAtTwo` behaviour or perhaps stop at any given `k` if desired?

Pascallio:
> @Pascallio Thanks for your contribution. It is very welcome! Maybe a stupid question/suggestion. We try to keep the dependencies as low as possible (Suggests contains already a large list of dependencies). Currently we successfully avoid `Rcpp` as direct dependency (some packages in Suggests of course already depend on `Rcpp`). It would be great if you ensure that `Rcpp` would end in `Suggests` and not in `Depends` or `Imports` or (implement this whole stuff in pure C).

Thank you @sgibb!  I fully understand the concern about dependencies. I don't have much experience with Rcpp, so how do I ensure it would end up in `Suggests`? Does that mean I should replace the Rcpp `Range` and change NumericVectors with for instance std::vector? 

As for implementing it in pure C, I have no experience with C so it might take me a while to submit a pure C implementation but I'm willing to give it a shot when I have some spare time.

lgatto:
Not sure if this PR is still relevant, but I would like to see a unit test, that compares the R- and C-level results.

PR #107, opened 2022-12-21
Add function to sum intensities with appropriate handling of NA values

jorainer:
This PR adds the `sumi` function that sums intensities. The difference to the base `sum(..., na.rm = TRUE)` is:

- the function returns `NA` if `lenght(x) == 0` (in contrast to `0` which is returned by `sum`).
- the function returns `NA` if all values in `x` are `NA` (again, in contrast to `sum` that returns `0` in these cases.


jorainer:
What puzzles me a bit is the performance of the function. Although it does the sum in `C` it's not really faster than the `sum` function (that does also type conversions etc):

```r
sumir <- function(x) {
    if (all(is.na(x)))
        return(NA_real_)
    sum(x, na.rm = TRUE)
}

library(MsCoreUtils)
library(microbenchmark)

set.seed(123)
x <- rnorm(10)

microbenchmark(
sumi(x),
sumir(x),
sum(x, na.rm = TRUE)
)
Unit: nanoseconds
                 expr min    lq     mean median    uq     max neval cld
              sumi(x) 297 321.5   532.11  341.5 363.5   17938   100   a
             sumir(x) 549 581.5 93749.98  603.5 680.5 9286872   100   a
 sum(x, na.rm = TRUE) 146 167.5   197.18  173.5 183.0     886   100   a

x <- rnorm(1000)
microbenchmark(
sumi(x),
sumir(x),
sum(x, na.rm = TRUE)
)
Unit: nanoseconds
                 expr  min     lq    mean median     uq   max neval cld
              sumi(x) 2357 2406.5 2466.38 2429.0 2445.5  6298   100   b
             sumir(x) 1829 1940.0 2231.03 1994.5 2076.0 15983   100   b
 sum(x, na.rm = TRUE)  848  871.5  944.02  883.0  899.5  3689   100  a 
```

... maybe @sgibb you can do something with your C-fu?

sgibb:
Mh, I would have done it nearly identical. Did you find the C source code of R's `sum`?

jorainer:
I guess it's [this here](https://github.com/wch/r-source/blob/trunk/src/main/summary.c#L152:L170). but that doesn't look so different to me?

lgatto:
Fine by me. 

A few questions:
- do you often have zero-length intensities or NA-only intensities?
- given that (for now at least) `sumi` is slower, are the benefits still worth it?

jorainer:
@lgatto 

> do you often have zero-length intensities or NA-only intensities?

actually yes. This happens frequently when LC-MS data gets filtered by m/z and rt (when extracting an EIC/chromatogram).

lgatto (approved):

Issue #108, opened 2023-01-04
bin: y argument must be an ordered vector


RogerGinBer:
Hi there!
I was looking into the function `bin` for calculating ion-mobility mobilograms (https://github.com/sneumann/xcms/pull/647) and, while experimenting, it seems that the argument `y` **must** be an increasingly ordered vector (otherwise it creates a silent bug). I would suggest adding a check for ordered data and/or documenting this requirement in the docs.

I'm guessing that for its original intended purpose (binning mz values) this was not an issue, as mz values are _always_ ordered in a `Spectra` object.

Here's a reproducible example of the problem:
```r
 xvec <- 1:20
 yvec <- c(41:50, 1:10)
 res_bad <- MsCoreUtils::bin(x = xvec, y=yvec)  ## Incorrect result
 res_good <- MsCoreUtils::bin(x = xvec[order(yvec)], y=sort(yvec)) ## Correct result
 plot(yvec, xvec)
 plot(res_bad$mids, res_bad$x)
 plot(res_good$mids, res_good$x)

## Vector relative ordering shouldn't matter when binning, yet in here the output$x is reversed
identical(MsCoreUtils::bin(x = 1:10, y=1:10),
               MsCoreUtils::bin(x = rev(1:10), y=rev(1:10))) ## FALSE, but it should be TRUE
```

jorainer:
Great catch @RogerGinBer ! May I suggest you make a PR (which includes a unit test) for this?

sgibb:
@RogerGinBer thanks for the catch and the report. Unfortunately I don't believe the solution is as easy as simply reordering `x` and `y`.

`bin(1:3, 1:3)` should only be identical to `bin(3:1, 3:1)` if the `breaks` are identical (which is `TRUE` for the default argument).

`bin(1:3, 1:3, breaks = 1:3)` should be different from `bin(1:3, 1:3, breaks = 3:1)` (which currently don't work).

I am wondering if we should allow unsorted inputs at all. Your current PR #110 silently reorders `x` and `y` which could yield unexpected errors because as user you could expect that `bin(1:3, 1:3) == rev(bin(1:3, rev(1:3)))`.

Currently I am thinking that we should check that `y` and `breaks` are sorted and throw an error if not. @jorainer maybe we should introduce a `.check` argument to skip these (as we did in `closest`) to not slow down the function in `Spectra` methods.

RogerGinBer:
[...]

PR #110, opened 2023-01-13
fix: check for y vector order in `bin`

RogerGinBer:
Add a check to `bin` to ensure y is ordered. #108
Update related documentation.
Add unit tests to ensure the fix works.

### DIFF
--- a/R/binning.R
+++ b/R/binning.R
@@ -8,9 +8,9 @@
 #'
 #' @param x `numeric` with the values that should be aggregated/binned.
 #'
-#' @param y `numeric` with same length than `x` with values to be used for
-#'     the binning. Ideally, `y` should be increasingly sorted, but it's not 
-#'     mandatory.
+#' @param y `numeric` with same length than `x` with values to be used for the
+#'   binning. `y` **must** be increasingly sorted, or else an error will be
+#'   thrown.
 #'
 #' @param size `numeric(1)` with the size of a bin.
 #'
@@ -24,6 +24,10 @@
 #'     `returnMids = FALSE` might be useful if the breaks are defined before
 #'     hand and binning needs to be performed on a large set of values (i.e.
 #'     within a loop for multiple pairs of `x` and `y` values).
+#'     
+#' @param .check `logical(1)` whether to check that `y` is an ordered vector.
+#'     Setting `.check = FALSE` will improve performance, provided you are sure
+#'     that `y` is always ordered.
 #'
 #' @return
 #'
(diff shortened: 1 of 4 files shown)


R/binning.R
```diff
@@ -9,7 +9,8 @@
 #' @param x `numeric` with the values that should be aggregated/binned.
 #'
 #' @param y `numeric` with same length than `x` with values to be used for
-#'     the binning.
+#'     the binning. Ideally, `y` should be increasingly sorted, but it's not 
+#'     mandatory.
```
sgibb, line 13: I am somehow against silently reordering `y`. Please see https://github.com/rformassspectrometry/MsCoreUtils/issues/108#issuecomment-1382383808

tests/testthat/test_binning.R
```diff
@@ -48,6 +48,16 @@ test_that("bin works", {
     expect_true(is.numeric(res_2))
     expect_equal(res$x, res_2)
 
+    ## Ensure y order does not matter
+    ## Issue #108 https://github.com/rformassspectrometry/MsCoreUtils/issues/108
+    expect_false(is.unsorted(xs))
```
sgibb, line 53: No need to check that base R functions are working. 
```suggestion
```

tests/testthat/test_binning.R
```diff
@@ -48,6 +48,16 @@ test_that("bin works", {
...
+    expect_equal(bin(vals, xs), bin(rev(vals), rev(xs)))
+    xs_unsorted <- sample(xs, length(xs))
+    expect_true(is.unsorted(xs_unsorted))
```
sgibb, line 56:
```suggestion
```

tests/testthat/test_binning.R
```diff
@@ -48,6 +48,16 @@ test_that("bin works", {
     expect_true(is.numeric(res_2))
     expect_equal(res$x, res_2)
 
+    ## Ensure y order does not matter
```
sgibb, line 51: I know it is done above as well but why we need such a "complex" unit test if I could simply test `bin(1:3, 1:3)` vs `bin(3:1, 3:1)`.

Unit tests should be as easy as possible and should require nearly no runtime (I know we are not good at both in all RforMassSpectrometry packages).

sgibb (changes requested):
Thanks for the effort of creating a PR. While @jorainer asked for it I like to discuss first if we really want to reorder `y` in #108. If we really agree that it should be done, we have to ensure that `breaks` are ordered as well.

R/binning.R
lgatto: I'm with @sgibb here and advise against silently re-ordering.

jorainer:
Agree with @sgibb . I would keep this function simple and throw an error if `x` is not ordered. This test could be made conditional with the `.check` parameter (default `.check = TRUE`) which would allow skipping it for data where it is known that the values are sorted.

@RogerGinBer , can you please also describe this requirement in the documentation? This function is supposed to allow binning of m/z values of a single spectrum and these should be increasingly ordered by definition.

R/binning.R
jorainer: Yes, agree. Please no silent re-ordering (see also https://github.com/rformassspectrometry/MsCoreUtils/issues/108#issuecomment-1400453934)

RogerGinBer:
Hi there, I've removed the silent reordering, created a `.check` param (default `TRUE`) to throw an error if `y` is unordered, and updated tests and docs accordingly. I think that should be all, let me know if we need anything else :+1:

jorainer (changes requested):
Thanks Roger @RogerGinBer for the PR! Only thing I would ask you to do it to add you as a contributor (`"ctb"`) to the `DESCRIPTION`.

Thanks!

jorainer (approved):

jorainer:
Thanks @RogerGinBer ! I'll merge now and push to BioC

Issue #111, opened 2023-02-07
Add a function to calculate the max intensity


jorainer:
Using `max` to determine the maximal intensity can have some issues, especially when chromatographic data is extracted and the spectra are thus subset by m/z first: a simple `max(i, na.rm = TRUE)` will return `-Inf` if there are no non-missing values. A function to determine the maximal intensity of a spectrum should:

- return the maximum removing any missing values
- return `NA_real_` if no non-missing values are present.

This is similar to the recently added `sumi` function.

A simple implementation could thus be:

```r
maxi <- function(x, na.rm = TRUE) {
    nna <- !is.na(x)
    if (any(nna))
        max(x[nna])
    else NA_real_
}
```

a C implementation might be more efficient here.

jorainer:
And again, it's surprising how efficient the base R `max` function actually is. The `maxi` C implementation is fast, but `max` is still faster - especially for larger `x`.

Some benchmarking:

```r
library(MsCoreUtils)
library(microbenchmark)

maxir <- function(x) {
    res <- max(x, na.rm = TRUE)
    if (is.finite(res))
        res
    else NA_real_
}
```

small `x` (length 10):

```r
x <- rnorm(10, mean = 100)
x[4] <- NA
microbenchmark(
    max(x, na.rm = TRUE),
    maxi(x),
    maxir(x)
)
Unit: nanoseconds
                 expr min  lq   mean median    uq  max neval cld
 max(x, na.rm = TRUE) 137 158 200.62  165.5 231.0  582   100 a  
              maxi(x) 293 306 362.34  320.0 335.5 3459   100  b 
             maxir(x) 424 444 527.56  460.0 518.5 4326   100   c
```

intermediate `x` (length 100):

```r
x <- rnorm(100, mean = 100)
x[4] <- NA
microbenchmark(
    max(x, na.rm = TRUE),
    maxi(x),
    maxir(x)
)
Unit: nanoseconds
                 expr min    lq   mean median    uq  max neval cld
 max(x, na.rm = TRUE) 208 226.0 274.87  231.0 236.5 1754   100 a  
              maxi(x) 601 641.5 718.98  665.5 680.5 4950   100   c
             maxir(x) 485 500.0 572.84  520.0 569.5 3129   100  b 
```

so, the R version becomes slightly faster. Even more so with larger `x`:

```r
x <- rnorm(1000, mean = 100)
x[4] <- NA
microbenchmark(
[...]

PR #112, opened 2023-02-08
feat: add function `maxi`

jorainer:
- Add `maxi` function to determine the maximum of a `numeric` returning `NA_real_` instead of `-Inf` if all elements are `NA` or if length of `x` is zero. Issue #111 .


jorainer:
See also issue #111 for more description and a benchmark.

jorainer:
@sgibb, suggestions and (even better) performance improvements highly welcome - but I simply can't see where I could eventually squeeze some more nanoseconds from the C implementation...

src/maxi.c
```diff
@@ -0,0 +1,29 @@
...
+    SEXP r;
+    const R_len_t n = XLENGTH(x);
+    PROTECT(r=allocVector(REALSXP, 1));
+    double* rp = REAL(r);
+    double* xp = REAL(x);
+    *rp = NA_REAL;
+
+    for (int i = 0; i < n; i++) {
+        if (!ISNA(xp[i])) {
+            if (ISNA(*rp) || (!ISNA(*rp) && xp[i] > *rp))
```
sgibb, line 22: You could reduce the number of comparison inside the loop to reduce the runtime.
1. Set `rp` to `-Inf` to avoid test for `NA` inside the loop and check for finite outside the loop.
2. No need to test `xp` for `NA` because `NA > *rp` is always false.

e.g.
``` r
library(microbenchmark)
library(collapse)
#> collapse 1.9.0, see ?`collapse-package` or ?`collapse-documentation`
#> 
#> Attaching package: 'collapse'
#> 
#>     D
library(inline)
devtools::load_all()
#> ℹ Loading MsCoreUtils

maxir <- function(x) {
    res <- max(x, na.rm = TRUE)
    if (is.finite(res))
        res
    else NA_real_
}

maxi2 <- cfunction(c(x = "double"), "
  if (!isReal(x))
    x = coerceVector(x, REALSXP);
  double* xp = REAL(x);
  const R_xlen_t n = XLENGTH(x);

  SEXP r = PROTECT(allocVector(REALSXP, 1));
  double* rp = REAL(r);
  *rp = R_NegInf;

  for (R_xlen_t i = n; i >= 0; i--)
    if (xp[i] > *rp)
        *rp = xp[i];

  if (!(R_FINITE(*rp)))
    *rp = NA_REAL;

  UNPROTECT(1);
  return(r);
")

x <- rnorm(10, mean = 100)
x[4] <- NA
microbenchmark(
    max(x, na.rm = TRUE),
    maxi(x),
    maxi2(x),
    maxir(x),
    check = "identical"
)
#> Unit: nanoseconds
#>                  expr  min     lq     mean median     uq     max neval cld
#>  max(x, na.rm = TRUE)  413  463.0   578.59  476.5  558.0    3684   100   a
#>               maxi(x) 1172 1234.5  1491.55 1275.0 1372.5   11934   100   a
#>              maxi2(x)  656  695.5   822.53  725.0  794.0    4956   100   a
#>              maxir(x) 1201 1245.0 37377.27 1289.5 1419.0 3595709   100   a

x <- rnorm(1000, mean = 100)
x[4] <- NA
microbenchmark(
    max(x, na.rm = TRUE),
    maxi(x),
    maxi2(x),
    maxir(x),
    check = "identical"
)
#> Unit: microseconds
#>                  expr    min     lq     mean  median      uq    max neval  cld
#>  max(x, na.rm = TRUE)  2.611  2.700  2.80259  2.7860  2.8605  3.867   100  b  
#>               maxi(x) 11.940 12.101 12.78183 12.6715 12.8435 23.387   100    d
#>              maxi2(x)  1.802  1.897  2.21136  2.0020  2.0830 16.639   100 a   
#>              maxir(x)  3.396  3.547  3.76781  3.6620  3.7415  9.897   100   c
```

<sup>Created on 2023-02-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/maxi.R
```diff
@@ -0,0 +1,45 @@
+#' @title Maximum MS Intensity Value
+#'
+#' @description
+#'
+#' `maxi` determines the maximum or mass spectrometry intensity values, e.g.
+#' from a spectrum or chromatogram. In contrast to the base R [max()] function
+#' this function returns `NA_real_` if all intensity values are `NA` or if
+#' `length(x)` is 0 (the base R `max` function returns `-Inf` in these cases).
+#'
+#' @param x `numeric` with intensity values to be summed up. Will be coerced
```
sgibb, line 10: *summed up*? (residual of `sumi`?)

R/maxi.R
```diff
@@ -0,0 +1,45 @@
...
+#'     to `numeric` using `as.double`.
+#'
+#' @return `numeric(1)` representing the sum of values in `x`. Always returns
```
sgibb, line 13:
```suggestion
#' @return `numeric(1)` representing the maximum of values in `x`. Always returns
```

sgibb (changes requested):
Just a few suggestions to slightly speed up the whole thing.

src/MsCoreUtils.h
```diff
@@ -27,4 +27,7 @@ extern SEXP _MsCoreUtils_imp_neighbour_avg(SEXP, SEXP);
 
 extern SEXP C_sumi(SEXP);
 
+extern SEXP C_maxi(SEXP);
+extern SEXP C_maxi2(SEXP);
```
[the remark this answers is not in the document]
jorainer: ups - yes, sorry. forgot to remove.

jorainer:
Thanks Sebastian! That updates in the C code rock! the only thing I changed is to loop from 0 to n instead of n to 0 - otherwise `maxi(numeric())` was returning `-Inf` instead of `NA_real_`.

I've addressed all comments - please have another look.

R/maxi.R
jorainer: thanks. updated.

src/maxi.c
jorainer: impressive @sgibb ! Only thing is that `maxi2(numeric())` will return `-Inf` instead of `NA_real_`. When looping from 0 to n instead it works (I guess the idea was that `x` was an increasingly sorted vector such as m/z are)

src/MsCoreUtils.h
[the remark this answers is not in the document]
jorainer: done

sgibb (approved):

PR #114, opened 2023-03-21
Master

lgatto:


jorainer (approved):
Looks good to me, thanks Laurent. I thought `requireNamespace` throws an error if the package is not avaialbe, but apparently not. We might need to fix this also in other packages.

PR #115, opened 2023-05-23
Adding the functions to calculate entropy

mar-garcia:


R/entropy.R
```diff
@@ -0,0 +1,40 @@
...
+#'
+#' @examples
+#' spectrum <- rbind(c(41.04, 37.16), c(69.07, 66.83), c(86.1, 999.0))
+#'
+#' entropy_spectral(spectrum[,2])
+#' entropy_normalized(spectrum[,2])
+#' 
+entropy_spectral <- function(x){
+  y <- x / sum(x)
+  z <- y*log(y)
```
jorainer, line 34: you could avoid creating the additional variable `z` here by simply using `-sum(y * log(y))` (note also the white space between `y` and `log`)

Also, no need to use `return` in an R function - the last value will always returned, there is no need for additional `return` function call.

jorainer (changes requested):
Thanks Mar @mar-garcia for these very nice functions!

Please address the comments and in addition:

- please add unit tests (in a file *tests/test_entropy.R* - see maybe [this](https://github.com/rformassspectrometry/MsCoreUtils/blob/main/tests/testthat/test_ppm.R) for an example.
- please add yourself as a contributor (ctb) to the DESCRIPTION
- please increase the version number (the last number 1.11.6 -> 1.11.7) in DESCRIPTION
- please add an entry to the NEWS.md describing the update/addition.

R/entropy.R
```diff
@@ -0,0 +1,40 @@
+#' @title Spectral entropy
+#'
+#' @description
+#'
+#' These functions allows to calculate entropy measurements of an MS/MS  
+#' spectrum based on the metrics suggested by 
+#' Li et al. (https://doi.org/10.1038/s41592-021-01331-z).
+#'
+#' @param x `numeric`, intensities of the fragment ions value(s).
```
lgatto, line 9: the "value(s)" can be dropped

lgatto (changes requested):
- I'll leave it to Johannes to provide more insight on the value of the metric, but are there any plans to also implement the entropy distance? 
- New functions should come with unit tests. I see that there are simple cases that are presented in the paper that could be implemented.

R/entropy.R
```diff
@@ -2,11 +2,19 @@
...
+#' of an spectra.
+#' MassBank of North America (MoNA) defines spectra entropy as the intensity 
+#' weighted spectral peak number 
+#' (https://mona.fiehnlab.ucdavis.edu/documentation/entropy. 
+#' Additionally it is suggested to label as low-quality spectrums the ones 
+#' that have a normalized entropy higher than 0.8 and spectral entropy 
+#' higher than 3. 
 #'
-#' @param x `numeric`, intensities of the fragment ions value(s).
+#' @param x `numeric`, intensities of the fragment ions(s).
```
lgatto, line 17: just ions

jorainer:
Regarding the similarity function: yes, the entropy-based similarity function will be added in a second step.

jorainer (approved):
Thanks Mar!

PR #121, opened 2024-01-15
Create a sequence of values with increasing differences between values

jorainer:
This PR adds a simple function `breaks_ppm` that calculates a sequence of values with increasing difference between pairs of values. This is in contrast to `seq` that creates a sequence of values with a constant difference between values.

A use case of this function would be e.g. for binning of peaks in mass spectra with bin sizes that increase with the m/z. The current default (that simply uses `seq` and thus a constant bin size) does not properly account for m/z dependent errors of (TOF) MS instruments.

I'm not super-happy with the name of the function (`breaks_ppm`), thus, happy for suggestions. Also, the function could also (easily) be implemented in C to improve performance.


R/binning.R
```diff
@@ -98,3 +99,62 @@ bin <- function(x, y, size = 1,
...
+#'
+#' @examples
+#'
+#' res <- breaks_ppm(20, 50, by = 1, ppm = 50)
+#' res
+#'
+#' ## difference between the values increases (by ppm)
+#' diff(res)
+breaks_ppm <- function(from = 1, to = 1, by = 1, ppm = 0) {
+    ## Maybe implement in C for higher performance.
```
sgibb, line 148: I would remove this comment. We can do it later if this function is really a bottle neck. But I assume it wouldn't be called very often.

R/binning.R
```diff
@@ -98,3 +99,62 @@ bin <- function(x, y, size = 1,
...
+    l <- ceiling((to - from + 1) / by)
+    res <- rep(NA_real_, l)
+    res[1L] <- a <- from
+    i <- 2L
+    while(a < to) {
+        a <- a + by
+        a <- a + ppm(a, ppm)
+        res[i] <- a
+        i <- i + 1L
+    }
```
sgibb, line 158: I would take the `ppm` calculation out of the loop:
```suggestion    
    p <- 1 + ppm(1, ppm)
    while(a < to) {
        a <- (a + by) * p
        res[i] <- a
        i <- i + 1L
    }
```
```r
library("microbenchmark")
to <- 1e3
by <- 0.5
p <- 50
microbenchmark(
    breaks_ppm(1, to, by, p),
    breaks_ppm2(1, to, by, p),
    check = "equal"
)
# Unit: microseconds
#                       expr      min       lq      mean  median       uq
#   breaks_ppm(1, to, by, p) 1447.536 1467.647 1541.1652 1518.72 1540.959
#  breaks_ppm2(1, to, by, p)  204.707  206.529  209.0002  207.35  209.256
#       max neval
#  2491.126   100
#   235.823   100
```

sgibb (changes requested):
Thanks, I have a small suggestion.

R/binning.R
jorainer: great suggestion! Thanks Sebastian!

sgibb:
@jorainer I already merged this PR but a version bump and NEWS.md entry is missing. Could you add this please?

PR #122, opened 2024-02-15
small correction on the definition of "ndotproduct"

mar-garcia:
I guess here it should be written "m > 0" instead of "n >0" because if I'm not wrong it is still talking about m/z weights and they are defined by "m" not "n" ("n" is for intensity weights)...


lgatto:
Indeed, I think you are right. @jorainer, could you confirm.

jorainer:
Yes, @lgatto , I confirm.

jorainer (approved):
Thanks @mar-garcia for the fix.

PR #123, opened 2024-03-18
Addition of function `force_sorted()` 

philouail:
Here I propose the function `force_sorted()` which takes as an input a vector of numeric values and interpolate the values that are not increasing. 

I wrote this function specifically for a retention alignment method that is being developed in xcms and at some point i had to deal with unsorted/non-increasing retention times values. However, I believe it could be used in the future or by user for some other reasons. 

Also, would welcome inputs on the name, as the function do no `sort()` the values per se, but we do check for the results by checking `is.unsorted()`. 

I also thought about a possible extra parameter `keep.na = TRUE` which would (if `FALSE`) allow the user to decide if they want the NA values to be adjusted too or not. 

Welcoming feedback :)


R/force_sorted.R
```diff
@@ -0,0 +1,51 @@
...
+#'
+#' @examples
+#' x <- c(NA, NA, NA, 1.2, 1.1, 1.14, 1.2, 1.3, NA, 1.04, 1.4, 1.6, NA, NA)
+#' sorted_rtime <- force_sorted(x)
+#' is.unsorted(x, na.rm = TRUE)
+#'
+#' @export
+#'
+#' @rdname force_sorted
+force_sorted <- function(x){
```
lgatto, line 22: Need a space between `)` and `{`.

R/force_sorted.R
```diff
@@ -0,0 +1,51 @@
...
+        stop("'x' needs to be numeric or integer")
+    nna_idx <- which(!is.na(x))
+    vec_temp <- x[nna_idx]
+
+    while (any(diff(vec_temp) < 0)) {
+        idx <- which.max(diff(vec_temp) < 0)
+        # Find next biggest value
+        next_idx <- which(vec_temp > vec_temp[idx])[1L]
+
+        if (is.na(next_idx)){
```
lgatto, line 34: Need a space between `)` and `{.

R/force_sorted.R
```diff
@@ -0,0 +1,51 @@
...
+            l <- idx:length(vec_temp)
+            vec_temp[l] <- seq(vec_temp[idx], by = 0.000001,
+                               length.out = length(l))
+            warning("Found decreasing values at the end of vector, ",
+                    "interpolation not possible. Replacing values. See help ",
+                    "for more details")
+            break
```
lgatto, line 41: What is the output supposed to be after the warning and the `break`? There's not detail in the manual, even though the message tells me to read it.

R/force_sorted.R
```diff
@@ -0,0 +1,51 @@
+#' @title Forcing a numeric vector into a monotonously increasing sequence.
+#'
+#' @description
+#' This function performs interpolation on the non-increasing parts of an
+#' input vector to ensure its values are continuously increasing.
```
lgatto, line 5: May be be consistent and use monotonously (as in the title) or continuously?

R/force_sorted.R
lgatto: Also, I don't get the warning message. It claims that interpolation isn't possible, even though the output is sorted despite the warning message:

```
> is.unsorted(force_sorted(c(1, 2, 1.5, 2)))
[1] FALSE
Warning message:
In force_sorted(c(1, 2, 1.5, 2)) :
  Found decreasing values at the end of vector, interpolation not possible. Replacing values. See help for more details
```

lgatto: And I see that the `by` is hard-coded to 0.000001. Is there a justification for that value? It could be a function argument?

lgatto (changes requested):

R/force_sorted.R
philouail: Ah yes I will rewrite the warning to be more clear. 
Even if the interpolation for the last value is not possible the values are increased by 0.0001.
 It make sense for our case but I guess for some user it is not the case, so as you said the `by = 0.0001` could be a function argument !

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # MsCoreUtils 1.15
 
+## MsCoreUtils 1.15.5
+
+- Add function `force_sorted()` to adjust a vector to ensure increasing/sorted
```
jorainer, line 5: maybe specify here that: *to adjust a numeric vector to ensure*.

R/force_sorted.R
```diff
@@ -0,0 +1,71 @@
+#' @title Forcing a numeric vector into a monotonously increasing sequence.
+#'
+#' @description
+#' This function performs interpolation on the non-increasing parts of an
```
jorainer, line 4: also here better to be clear: *... parts of a numeric input vector ...*.

R/force_sorted.R
```diff
@@ -0,0 +1,71 @@
...
+#' input vector to ensure its values are monotonously increasing.
+#' If the values are non-increasing at the end of the vector, these values
+#' will be monotonously increasing by the value defined with parameter `by `
```
jorainer, line 7: maybe better here *... these values will be replaced by a sequence of numeric values, starting from the last increasing value in the input vector, and increasing by a very small value, which can be defined with parameter `by`. *

R/force_sorted.R
```diff
@@ -0,0 +1,71 @@
...
+#' from the last increasing value.
+#'
+#' @param x `numeric` vector.
+#'
+#' @param by `numeric(1)` value that will determine the monotonous increase in
+#'        case the values at the end of the vector are non-increasing and
+#'        therefore interpolation would not be possible. Default
```
jorainer, line 14: *Default* -> *Defaults*

R/force_sorted.R
```diff
@@ -0,0 +1,71 @@
...
+#'        to `by = .Machine$double.eps` which is the smallest positive
+#'        floating-point number x such that 1 + x != 1.
+#'
+#' @return A vector with continuously increasing values.
+#'
+#' @note
+#' NA values will not be replaced.
```
jorainer, line 21: Maybe also be a bit more clear here: *NA values in `x` will not be replaced and returned as is.*.

src/MsCoreUtils.dll
jorainer, line 1: can you please remove this file? it's also strange that it is added, so please add also dll files to the .gitignore file.

jorainer (changes requested):
Thanks Phili for the contribution. I'm generally OK, only some minor fixes in the documentation.

jorainer:
Maybe also to give a bit more of a context to this function: in the alignment (retention time adjustment) of LC-MS data we sometimes observe that the order of adjusted retention times of spectra is no longer ordered - which in fact should not happen, because spectra should always be ordered as they were acquired in the LC-MS run - and this is ordered by retention time. So, in cases were some spectra have out-of-order adjusted retention times we want to *force* them into order again.

At present, this code is used in `xcms`, but will then eventually also be needed in the `Chromatograms` package - hence adding it here to `MsCoreUtils`.

R/force_sorted.R
```diff
@@ -0,0 +1,72 @@
...
+#' is.unsorted(x, na.rm = TRUE)
+#'
+#' ## Vector non increasing at the end
+#' x <- c(1, 2, 1.5, 2)
+#' sorted_rtime <- force_sorted(x, by = 0.1)
+#' is.unsorted(x, na.rm = TRUE)
+#'
+#' ## We can see the values were not interpolated but rather replaced by the
+#' ## last increasing value `2` and increasing by 0.1.
+#' sorted_vec
```
lgatto, line 36: I think this should be `sorted_rtime`. Also could you use another name, like x2 or y, to keep it general. Here, I guess rtime refers to retention time, but this isn't mentioned in the man page (and there's no need to), so it is a bit of an obscure name.

R/force_sorted.R
```diff
@@ -0,0 +1,72 @@
...
+        next_idx <- which(vec_temp > vec_temp[idx])[1L]
+
+        if (is.na(next_idx)) {
+            l <- idx:length(vec_temp)
+            vec_temp[l] <- seq(vec_temp[idx], by = by,
+                               length.out = length(l))
+            warning("Found decreasing values at the end of the vector. ",
+                    "Interpolation is not possible in this region. Instead, ",
+                    "replacing these values with a sequence that starts from ",
+                    "the last increasing value and increments by  ", by,
```
lgatto, line 60: There are two spaces after by.

lgatto (changes requested):

lgatto (approved):

lgatto:
Thank you for the contribution!

jorainer (approved):
Thanks Phili! green lights also from my side. Will merge and push to BioC.

Thanks!

Issue #125, opened 2024-03-18
Partial argument matching


lgatto:
In [this line](https://github.com/rformassspectrometry/MsCoreUtils/blob/5a83ebe74d2a8e8d945e542f94cf5593db5101d5/R/rbindFill.R#L61), we have partial matching.

It should be `seq(along.with = l)`.

PR #127, opened 2024-03-18
fix partial arg match - close issue #125

lgatto:


R/rbindFill.R
```diff
@@ -58,7 +58,7 @@ rbindFill <- function(...) {
     allnms <- unique(names(allcl))
     allcl <- allcl[allnms]
 
-    for (i in seq(along=l)) {
+    for (i in seq(along.with = l)) {
```
sgibb, line 61: We use `seq_along` everywhere else.

```suggestion
    for (i in seq_along(l)) {
```

sgibb (approved):
Nice catch.

lgatto:
These warnings start to pop up since I set the following in my `.Profile`

```
options(warnPartialMatchArgs = TRUE,
        warnPartialMatchDollar = TRUE,
        warnPartialMatchAttr = TRUE)
```

jorainer (approved):
Thanks! feel free to merge and push!

lgatto:
Pushed to Bioc.

PR #128, opened 2024-04-17
Add common_path function

jorainer:
Add a simple file utility function `common_path()` that allows to extract the *common* file path from a set of provided files. This works on both unix and Windows systems. I will be used to allow changing the base path to e.g. mzML files for `MsBackendMzR` backends (or similar) if they have been moved to a different folder.


R/file_utils.R
```diff
@@ -0,0 +1,57 @@
...
+        return(character())
+    sx <- strsplit(x, split = "(\\\\)|/")
+    minl <- min(lengths(sx))
+    cpath <- character()
+    for (i in seq_len(minl)) {
+        uvals <- unique(vapply(sx, `[`, character(1), i = i))
+        if (length(uvals) == 1L)
+            cpath <- c(cpath, uvals)
+    }
+    paste0(cpath, collapse = fsep)
```
sgibb, line 56: I thought `file.path` should be used here. But `do.call(file.path, as.list(cpath))` is much slower than `paste0`.

jorainer: I know, it's funny/unexpected. but `paste0` does the trick here, and there is no additional benefit from using `file.path`.

jorainer: Are you generally OK with this PR @sgibb - would like to merge this before the Bioc devel branch freeze.

sgibb (approved):

PR #129, opened 2024-10-20
fix: use R_Calloc/R_Free instead of Calloc/Free to reflect changes in R API

sgibb:
Use `R_Calloc` and `R_Free` instead of `Calloc` and `Free` in  `src/lowerConvexHull.c`, respectively, to reflect changes in the R API and fullfil `STRICT_R_HEADERS` check.


jorainer (approved):
Thanks Sebastian! I'll merge and push to BioC

PR #132, opened 2025-03-06
Add reduce function

jorainer:
Add a function `reduce()` that (similar to the `IRanges::reduce()` method) *reduces* overlapping numeric ranges to disjoint ranges. The function takes two numeric vectors as input (with the min/max (start/end) for each range) and returns a `list` with two numeric vectors with the *reduced* ranges.

This function will be helpful in cases such as filtering spectra by multiple retention time ranges, as it allows to reduce them first to non-overlapping ones.

This is a pure R implementation. Happy for anybody trying a C implementation (@sgibb ?) that is expected to be much faster.


sgibb:
@jorainer thanks for this PR. As requested I translated your R implementation into C (I didn't check if there are better algorithms or implementations).

jorainer (commented):
Thank you so much @sgibb ! This is indeed much faster than the R implementation. For me the code looks good - we just need someone else to formally approve the PR.

R/reduce.R
```diff
@@ -0,0 +1,50 @@
...
+#' thus it is suggested to specifically import it if used in another R package,
+#' or to call it with `MsCoreUtils::reduce()`.
+#'
+#' @param start `numeric` with the lower (start) values for each numeric range.
+#'
+#' @param end `numeric` with the upper (end) values for each numeric range.
+#'     Has to match the length of `start` and `all(start <= end)` has to be
+#'     `TRUE`.
+#'
+#' @param .check `logical(1)` whether input parameter checks should be
```
philouail, line 23: What checks does it do ?

philouail (approved):
It makes sense to me, I was a bit confused but the `@return` made it clear. I would maybe details a bit more the `.check` parameter. but otherwise it's approved on my side.

R/reduce.R
jorainer: Updated with the recent push

PR #147, opened 2026-04-14
fix: handle all type of errors

gabrieletome:


jorainer (changes requested):
Thanks @gabrieletome ! Can you please also add unit tests for other types of errors if possible to see if it works as expected?

Then:
- please add yourself to the `@author` field of the present function
- change the version of the package to 1.23.9 in the DESCRIPTION
- add this version to the NEWS.md file describing what was changed
- add yourself as a contributor ("ctb") to the DESCRIPTION

jorainer (approved):
Thanks Gabriele!

jorainer:
hm - the CI errors are a bit odd. I will try to check locally on the R/BioC devel

PR #148, opened 2026-04-19
Add one more function "group_mz_int" to group.R 

MuyaoXi9271:
The discussion can be check from 
https://github.com/rformassspectrometry/Spectra/pull/382


lgatto:
- What is the empty test.txt file? 
- Could you please add a unit test for the new functionality?

MuyaoXi9271 (commented):
This is the function that I would like to request

lgatto:
What is the empty test.txt file?

> This is the function that I would like to request

The function is in an R file. 

In addition to the code in the `group_mz_int` function, we need:

> Could you please add a unit test for the new functionality?

and documentation. 

It is also important to only modify the lines you change. Currently, code lines from `group` also show up in the diff. The only changes seem to be spaces, but this makes the review unnecessarily cumbersome.

lgatto:
Hi @MuyaoXi9271 - it is really important to also add unit tests to your PR (your can find those that are already implemented in the `tests/testthat` directory). This really helps to first crystallise your intentions, and then assure that future changes don't inadvertently break the that intended behaviour.

R/group.R
```diff
@@ -65,3 +65,56 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+    df_xy <- data.frame("mz" = x, "int" = y, "group_id" = rep(0, length(x)), "index" = 1:length(x))
+    group_id_zero <- which(df_xy$group_id == 0)
+    group_id <- 0
+    
+    
+    while (length(group_id_zero) != 0) {
+        group_id <- group_id + 1
+        id_int_max <- group_id_zero[which.max(df_xy[group_id_zero, "int"])]
+        
+        logi_id <- with(df_xy, abs(df_xy[group_id_zero, "mz"]-df_xy[id_int_max, "mz"])<=tolerance)
```
jorainer: I think this will break if `ppm` is provided. Then, `tolerance` has `length(x) - 1`, which will not be the same as the length of `group_id_zero`.

I would calculate the `tolerance` specifically for the currently evaluated max intensity (i.e., `id_int_max`):

```r
logi_id <- abs(x[group_id_zero] - x[id_int_max]) <= tolerance + ppm(x[id_int_max], ppm)
```

This should also be a bit faster, because we don't need to subset the `data.frame` (related to previous comment).

R/group.R
```diff
@@ -65,3 +65,56 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+        tolerance <- tolerance + ppm(x[-length(x)], ppm)
+    
+    df_xy <- data.frame("mz" = x, "int" = y, "group_id" = rep(0, length(x)), "index" = 1:length(x))
+    group_id_zero <- which(df_xy$group_id == 0)
+    group_id <- 0
+    
+    
+    while (length(group_id_zero) != 0) {
+        group_id <- group_id + 1
+        id_int_max <- group_id_zero[which.max(df_xy[group_id_zero, "int"])]
```
jorainer: I suggest avoiding to subset a `data.frame` by row. that is slow. Rather, I would suggest to change this into

```r
id_int__max <- which.max(y[group_id_zero])
```

Also, I don't see a reason to use a `data.frame` at all here. Better to use the m/z (`x`) and intensity (`y`) vectors directly. It's faster and requires no additional copies of the data.

R/group.R
```diff
@@ -65,3 +65,56 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+    #group m/z within tolerance based on the intensity decreasingly
+    if(is.unsorted(x)) {
+        idx <- order(x)
+        x <- x[idx]
+        y <- y[idx]
+    } else idx <- integer()
+    
+    tolerance <- tolerance + sqrt(.Machine$double.eps)
+    if (ppm > 0)
+        tolerance <- tolerance + ppm(x[-length(x)], ppm)
```
jorainer: I suggest to remove this and calculate the for each selected peak separately using `ppm()` (see comment below)

R/group.R
```diff
@@ -65,3 +65,56 @@ group <- function(x, tolerance = 0, ppm = 0) {
     res[idx] <- res
     res
 }
+
+
+#---- New lines for adding new function for grouping based on the corresponding intensity decreasingly --#
+group_mz_int <- function(x, y, max_num, tolerance = 0, ppm = 0) {
+    #group m/z within tolerance based on the intensity decreasingly
+    if(is.unsorted(x)) {
```
jorainer: I think this whole `if(is.unsorted(x))...` can be be removed, as your approach does not require ordered m/z values (if I understand it correctly).

R/group.R
```diff
@@ -65,3 +65,56 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+        idx <- order(x)
+        x <- x[idx]
+        y <- y[idx]
+    } else idx <- integer()
+    
+    tolerance <- tolerance + sqrt(.Machine$double.eps)
+    if (ppm > 0)
+        tolerance <- tolerance + ppm(x[-length(x)], ppm)
+    
+    df_xy <- data.frame("mz" = x, "int" = y, "group_id" = rep(0, length(x)), "index" = 1:length(x))
```
jorainer: no need to use a `data.frame`. the `x` and `y` vectors can be directly used, and the resulting group id could be defined as a vector too (e.g. `res <- rep(0, length(x)))`.

jorainer: This has also performance advantages, see comments below.

R/group.R
```diff
@@ -65,3 +65,56 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+            index_top_int <- tail(index_sel_int, max_num)
+            index_mimus1_int <- setdiff(index_sel_int, index_top_int)
+            
+            df_xy[int_index[index_top_int, "index"], "group_id"] <- rep(group_id, length(index_top_int))
+            df_xy[int_index[index_mimus1_int, "index"], "group_id"] <- rep(-1, length(index_mimus1_int))
+        } else {
+            df_xy[group_id_zero[logi_id], "group_id"] <- rep(group_id, sum(logi_id))
+        } 
+        
+        group_id_zero <- which(df_xy$group_id == 0)
```
jorainer: here of course, it would be `group_id_zero <- which(res == 0)` if `res` would be used instead of `df_xy`.

R/group.R
```diff
@@ -65,3 +65,56 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+    while (length(group_id_zero) != 0) {
+        group_id <- group_id + 1
+        id_int_max <- group_id_zero[which.max(df_xy[group_id_zero, "int"])]
+        
+        logi_id <- with(df_xy, abs(df_xy[group_id_zero, "mz"]-df_xy[id_int_max, "mz"])<=tolerance)
+        
+        if (sum(logi_id) == 0) {
+            df_xy[id_int_max, "group_id"] <- group_id
+            
+        } else if (sum(logi_id) > max_num) {
```
jorainer: I don't understand what this part is doing? Can you please explain @MuyaoXi9271 ?

R/group.R
```diff
@@ -65,3 +65,56 @@ group <- function(x, tolerance = 0, ppm = 0) {
     res[idx] <- res
     res
 }
+
+
+#---- New lines for adding new function for grouping based on the corresponding intensity decreasingly --#
```
jorainer, line 75: Documentation is missing here. Please add the documentation to describe what the function does, the individual parameters and what the result is. Importantly, describe what the properties of the grouping is. If I understand correctly, it groups peaks with an *m/z* smaller than `tolerance` + `ppm( ppm)` into a group, starting from the peaks with the highest intensity. All peaks within each group have then a similar m/z.

 **but** it could also be that other peaks, close to a group, have an m/z value for which the difference to the m/z value of any peak within the group is also < `tolerance` - just, its difference to m/z of the **highest intensity peak** of the group is > `tolerance`. Correct me if I'm wrong. Also, I don't think that's a problem, but it needs to be clearly explained in the documentation.

jorainer (changes requested):
Thanks @MuyaoXi9271 ! I have some questions and would request changes on your code.

Also, importantly, you're missing
- unit tests
- documentation

These should also be added (ideally, follow the documentation of the other `group` functions in this file and the unit tests in *tests/testthat/test_group.R*).

thanks

R/group.R
MuyaoXi9271: I have added the documentation for the new function and test file inside the folder of testthat as well. Then m/z with the second highest intensity would be picked up and similar ones would be found, etc

R/group.R
MuyaoXi9271: For the case like only one peak in spectrum, so only keep that peak as group 1

R/group.R
MuyaoXi9271: Done, please check latest uploaded function

R/group.R
MuyaoXi9271: Done, please check latest uploaded function

R/group.R
MuyaoXi9271: Done, please check latest uploaded function

R/group.R
MuyaoXi9271: Done, please check latest uploaded function

R/group.R
MuyaoXi9271: Done, please check latest uploaded function

R/group.R
MuyaoXi9271: Done, please check latest uploaded function

R/group.R
```diff
@@ -65,3 +65,148 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+#' ## Same on an unsorted vector
+#' x <- c(56, 56.012, 56.016, 56.004, 56.008, 56.02)
+#' y = c(52151, 58452, 458, 125584, 582, 57452)
+#' group_mz_int(x, y, max_num, tolerance = 0.005)
+#'
+#' ## the same three groups were made.
+
+
+
+group_mz_int <- function(x, y, max_num, tolerance = 0, ppm = 0) {
```
jorainer: I would suggest to set `max_num = Inf` - so the default would be to keep all peaks within the tolerance.

R/group.R
```diff
@@ -65,3 +65,148 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+    #group m/z within tolerance based on the intensity decreasingly
```
jorainer, line 142: Please remove all the commented-out code lines

R/group.R
```diff
@@ -65,3 +65,148 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+        tolerance <- tolerance + ppm(x[1:length(x)], ppm)
+    
+    
+    
+    #df_xy <- data.frame("mz" = x, "int" = y, "group_id" = rep(0, length(x)), "index" = 1:length(x))
+    
+    mz <- x
+    int <- y
+    group_ids <- rep(0, length(x))
+    index <- 1:length(x)
```
jorainer: please use `seq_along(x)` instead of `1:length(x)`: if `length(x)` is `0` you would otherwise get an `index = 1` and an out-of-bounds error.

R/group.R
```diff
@@ -65,3 +65,148 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+    #group m/z within tolerance based on the intensity decreasingly
+    # if(is.unsorted(x)) {
+    #     idx <- order(x)
+    #     x <- x[idx]
+    #     y <- y[idx]
+    # } else idx <- integer()
+    
+    tolerance <- rep(tolerance + sqrt(.Machine$double.eps), length(x))
+    if (ppm > 0)
+        tolerance <- tolerance + ppm(x[1:length(x)], ppm)
```
jorainer: the `x[1:length(x)]` can be replaced with `x` - no need for an additional subset operation.

R/group.R
```diff
@@ -65,3 +65,148 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+        
+        #logi_id <- with(df_xy, abs(df_xy[group_id_zero, "mz"]-df_xy[id_int_max, "mz"])<=tolerance[group_id_zero])
+        
+        logi_id <- abs(mz[group_id_zero]-mz[id_int_max])<=tolerance[group_id_zero]
+        
+        if (sum(logi_id) == 0) {
+            #df_xy[id_int_max, "group_id"] <- group_id
+            group_ids[id_int_max] <- group_id
+            
+        } else if (sum(logi_id) > max_num) {
```
jorainer: here I would use `if (is.finite(max_num) && sum(logi_id) > max_num)` - then the `sum(logi_id)` would only be calculated if `max_num` is set to a finite number.

R/group.R
```diff
@@ -65,3 +65,148 @@ group <- function(x, tolerance = 0, ppm = 0) {
...
+            
+            #int_index <- df_xy[group_id_zero[logi_id], c("int", "index")]
+            
+            int_logi <- int[group_id_zero[logi_id]]
+            index_logi <- index[group_id_zero[logi_id]]
+            
+            #index_sel_int <- order(int_index$int)
+            index_sel_int <- order(int_logi)
```
jorainer, line 177: I have some difficulties to understand WHY it is important to have only e.g. the 5 peaks with the highest intensity in the same group - even if there are other peaks with m/z below tolerance. why then not group the 5 peaks with the most similar m/z?

jorainer (changes requested):
Thanks @MuyaoXi9271 for the updates and the documentation. Can you please address my latest requests. And also, please add yourself as `"ctb"` to the DESCRIPTION.

jorainer:
And please add `#' @importFrom utils tail` to the documentation of the function to import the function and fix the failure of the unit tests.

R/group.R
MuyaoXi9271: I would like to get a sequence number from 1 to length(x). If `1:length(x)` is replaced with `seq_along(x)`, only one number will be obtained. 

I also test `group_mz_int(c(), c(), Inf, tolerance = 0, ppm = 0)` turns to be numeric(0)

R/group.R
MuyaoXi9271: To order peaks based on intensity is just the first step to look for the groups, which peaks are from several spectra shared with the same precursor ion. 

Initially, all peaks are given the same group number 0. Then the highest peak would be picked up after ordering peaks by intensity; similar masses within the threshold would be grouped, and the grouped peaks would be assigned the same group number. Again, the highest peak would be picked up from the remaining peaks still assigned the name group 0; its similar masses within the threshold would be given another group number. The steps above would be repeated until all peaks have their group number.

R/group.R
jorainer: that's strange - `seq_along(x)` should return a sequence from 1 to the length of `x` too:

```r
> seq_along(c(4, 3, 4, 1))
[1] 1 2 3 4
```

while 

```r
> seq_along(c())
integer(0)
```

but

```r
> 1:length(c())
[1] 1 0
```

using `seq_along(x)` is therefore safer than `1:length(x)`.

R/group.R
jorainer: yes, that's what I also understand from the code. I was just not sure why you implemented it in this way, because it could result in peak groups with overlapping m/z range? while if you ordered them by difference in m/z you would group the top x peaks with the most similar m/z. your approach groups the top x peaks with the most similar intensity (having a difference in their m/z which is within tolerance).

If that is really what you intended then it only has to be clearly described in the documentation.

jorainer: @MuyaoXi9271 , could you please confirm whether that was the way you wanted to group the peaks?

R/group.R
MuyaoXi9271: Yes, you are right. I changed that already

R/group.R
MuyaoXi9271: 1. I implemented this approach to avoid grouping m/z values above the threshold; meanwhile, higher-intensity m/z values play a more important role in the MS2 spectrum. Therefore, I order them by intensity
2. And I do not think the peak groups will overlap the m/z range
3. If the multiple scans for the same precursor behave similarly, the top x peaks with the most similar intensity are selected, and these top x peaks are from these x scans

I also modified a bit to make it clear in the documentation. Please let me know whether I explained it clearly or which part I need to provide more description

jorainer: Thanks for the explanation.

jorainer (changes requested):

jorainer (changes requested):
Thanks @MuyaoXi9271 . I'm OK with the code, documentation and unit tests. I will fix the missing things (import) after merging. Also, I will do some minor *cosmetic* changes to your code (remove excessive white spaces and ensure the style fits ours) **after** merging.

But before merging I would ask you to please add yourself as a contributor to the DESCRIPTION file following the style of e.g. https://github.com/rformassspectrometry/MsCoreUtils/blob/main/DESCRIPTION#L39-L41

Thanks!

jorainer:
or share your name and email address (and optional ORCID) here and I will add it.

MuyaoXi9271:
> Thanks @MuyaoXi9271 . I'm OK with the code, documentation and unit tests. I will fix the missing things (import) after merging. Also, I will do some minor _cosmetic_ changes to your code (remove excessive white spaces and ensure the style fits ours) **after** merging.
> 
> But before merging I would ask you to please add yourself as a contributor to the DESCRIPTION file following the style of e.g. https://github.com/rformassspectrometry/MsCoreUtils/blob/main/DESCRIPTION#L39-L41
> 
> Thanks!

Thanks for your approval and for updating some details to align the style with others. I have updated my information in the file DESCRIPTION. Please let me know what else you need from my side.

Best,
Muyao

jorainer (approved):
Thanks!

